Issue #8, opened 2025-07-20
llCollisionSprite


Martin-Pitt:
Double checking when going through the list of functions I notice that [per the wiki](https://wiki.secondlife.com/wiki/LlCollisionSprite) that `llCollisionSprite` is marked as broken since 2007.

If that's the case, should it be marked as effectively deprecated?

bennettgoble:
Yeah, according to [VWR-322](https://github.com/secondlife/jira-archive-internal/issues/134198) collision sprites were removed quite a long time ago. Let's mark it as deprecated.

Martin-Pitt:
There were more deprecations I found which were for remote data that were missed (event and constants). The tooltip was very clear on this so I made the commits ahead into #13. Separated out own PR as it introduces a minor schema change (first time event is marked as deprecated).

PR #10, opened 2025-07-20
Mark `llCollisionSprite` as deprecated

Martin-Pitt:
Per #8 the function has been broken for a long time


Martin-Pitt:
Added #13 as well, to extend the existing deprecations of remote data to the event (this is a schema change, first deprecated event?) and related remote data event constants (there have been deprecated constants before such as `ATTACH_LPEC`).

bennettgoble (approved):
LGTM, thanks!

Issue #11, opened 2025-07-21
`quaternion` type — omit from document generation?


Martin-Pitt:
It wasn't really worth doing but I did it anyways and had auto generated documents from the types and I spotted this [quaternion type](https://github.com/secondlife/lsl-definitions/blob/main/lsl_definitions.yaml#L11973-L11976) which I never knew about.

Since it's completely obsolete/deprecated and only useful for tooltips and syntax highlighting purposes it should probably be omitted from document generation e.g. marked with a `private: true`. Thoughts?
<br/>
<hr/>

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #11</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @Martin-Pitt</code> (replace <code>20</code> with the amount, and <code>@Martin-Pitt</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

PR #12, opened 2025-07-21
Mark `quaternion` type as private

Martin-Pitt:
Per #11 change to mark `quaternion` type as private to omit it from document generation


bennettgoble (approved):
LGTM, thanks!

PR #13, opened 2025-07-21
Mark `remote_data` event & related constants `REMOTE_DATA_*` as deprecated

Martin-Pitt:
More deprecations per #8. The tooltip and all related functions for remote data are already deprecated so this is a straightforward correction


bennettgoble (approved):
Ah, yes... I remember removing this service. We can do it... it just takes 13 or so years.

Issue #14, opened 2025-07-26
`LAND_SMALL_BRUSH` / `LAND_MEDIUM_BRUSH` /  `LAND_LARGE_BRUSH`


Martin-Pitt:
While investigating a related ground function (`llSetGroundTexture` has no documentation on the return value, still a mystery), I noticed a ridiculous situation with `llModifyLand`.

The small brush reads as:
```yaml
LAND_SMALL_BRUSH:
  tooltip: 'Use a small brush size.\nNOTE: This value is incorrect, a small brush
    should be 0.'
  type: integer
  value: 1
```
Same with the other brush constants

Is this a legacy error? If the LSL definition is really the source of truth — could this be amended and should it?

This issue is unimportant/trivial in the grand scheme of things and I was thinking of not bringing it up, but it does help settle how far reaching error fixes to the definition file can be and what the resolution process to issues could be as well as a case study. An alternative fix for example could be a new set of constants with the correct enumeration and deprecating the old constants.
<br/>
<hr/>

<details><summary>This repo is using Opire - what does it mean? 👇</summary><br/>💵 Everyone can add rewards for this issue commenting <code>/reward 100</code> (replace <code>100</code> with the amount).<br/>🕵️‍♂️ If someone starts working on this issue to earn the rewards, they can comment <code>/try</code> to let everyone know!<br/>🙌 And when they open the PR, they can comment <code>/claim #14</code> either in the PR description or in a PR's comment.<br/><br/>🪙 Also, everyone can tip any user commenting <code>/tip 20 @Martin-Pitt</code> (replace <code>20</code> with the amount, and <code>@Martin-Pitt</code> with the user to tip).<br/><br/>📖 If you want to learn more, check out our <a href="https://docs.opire.dev">documentation</a>.</details>

bennettgoble:
@HaroldCindy yeah I'm thinking we may want to document but not fix this problem as it could break content.

Martin-Pitt:
Leaving it as a documentation issue then

PR #16, opened 2025-07-31
llSetGroundTexture - no return value according to implementation

Martin-Pitt:
The function `llSetGroundTexture` does not actually return anything meaningful in the implementation.

According to [@HaroldCindy](https://github.com/secondlife/lsl-definitions/issues/14#issuecomment-3137546044) and Monty Linden (per SL discord #scripting) checking the code there were no return values.

If that's the case then we could tweak the definition to match with implementation.

Alternatively, this PR could be rejected and the implementation updated to match with estate management functions e.g. `llManageEstateAccess`:
> Returns a boolean (an integer) TRUE if the call was successful; FALSE if throttled, invalid action, invalid [...] or object owner is not allowed to manage the estate.

### DIFF
--- a/lsl_definitions.yaml
+++ b/lsl_definitions.yaml
@@ -10451,7 +10451,7 @@ functions:
         type: list
     energy: 10.0
     func-id: 558
-    return: integer
+    return: void
     sleep: 0.0
     tooltip: Changes terrain texture properties in the region.
   llSetHoverHeight:


HaroldCindy (changes requested):
This can't be changed to `void` because that would break API compatibility, a note should be added to explain that the retval has no meaning for now. We can update the note if / when that changes.

PR #17, opened 2025-08-06
Placeholder constant for llRequestAgentData

Rider-Linden:
Adding placeholder constant for llRequestAgentData


HaroldCindy:
@Rider-Linden Is this still needed?

PR #18, opened 2025-08-30
Deprecate `llSetPrimitiveParams` and `llSetLinkPrimitiveParams`

Martin-Pitt:


HaroldCindy:
@Rider-Linden @bennettgoble thoughts?

Rider-Linden (commented):
`llSetLinkPrimitiveParamsFast` does everything that both those functions do, and in fact all three go to the same code under the hood.  

If we pair this with a documentation update.  Something along the lines of:
> Use `llSetLinkPrimitiveParamsFast` rather than this function. 

Then I think it's a good idea.  (Does the lua compiler issue a warning for deprecated functions?)

HaroldCindy:
@Rider-Linden It doesn't, but there's no reason it couldn't, so long as there's no aliasing (i.e. you do `ll.SomeFunc()` rather than `local foo = ll; foo.SomeFunc()`

PR #28, opened 2025-11-15
Add Yaml Schema validation using yamale

WolfGangS:
This adds a schema document and a validation method for PR's to make sure that the `lsl_definitions.yml` follows the so far agreed schema.

It uses [Yamale](https://github.com/23andMe/Yamale) to perform the validation, which is installed by adding it as a requirement to the existing pyproject

There is also a small change to the config for the pre-commit action, as yamale uses multi-document yaml files for its validation method, so the pre-commits yaml check, needs to allow for this.

It also inclides 1 small change to the lsl_definitions.yml, to standardize a function, to match the rest, for the schema

### DIFF
--- a/lsl_definitions.yaml
+++ b/lsl_definitions.yaml
@@ -4738,6 +4738,7 @@ functions:
 #  <key>: is the LSL name for the function.
 #    energy: Script energy consumed by the function. (When in doubt just use 10.0)
 #    sleep: Additional sleep time imposed by the script after this function completes.
+#    mono-sleep: Sleep in the mono script runtime
 #    return: Return type of the function.
 #    arguments: An array of maps of maps describing the parameters to this function.
 #    tooltip: A brief description of this function. Map of maps is used to retain ordering.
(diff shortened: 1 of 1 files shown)


lsl_definitions.yaml
Martin-Pitt: - `linden-experience` is already commented, so that is a duplicate
- `mono-sleep` could be reordered below `sleep`
- `must-use` might make more sense next to `pure`

WolfGangS: Done

.pre-commit-config.yaml
```diff
@@ -8,5 +8,6 @@ repos:
     rev: v5.0.0
     hooks:
     -   id: check-yaml
+        args: [--allow-multiple-documents]
```
HaroldCindy: You mention that this is necessary because of something that yamale does, does yamale create json documents in the current directory that `ruff` then ends up picking up? Maybe better to not create those / create them in a working directory out of the source tree

WolfGangS: No, it's because of how the schema file for yamale works.

This line here https://github.com/secondlife/lsl-definitions/pull/28/files#diff-952095f3d61faa403ce145f93e4cf9887878a9a5f6cadcd0ce76db203ab7d6b7R7

That `---` is [yaml structure markup](https://yaml.org/spec/1.2.2/#22-structures) that means anything after is a "new " yaml document yamale uses the format to separate parts of the schema.

So the `check-yaml` step chokes on the schema file unless you allow multiple documents per file.

HaroldCindy (approved):

PR #30, opened 2025-12-03
Fix schema and it's validation

buildersbrewery:
This commit ...

- updates the workflow that produces test output so  that the YAML file is checked against the schema first
- uses JSON Schema's format instead of `yamale`'s so we can
  - use LSP with LSP-json to validate the schema
  - and use LSP with LSP-yaml to validate YAML code
- adds an option to validate the YAML file locally against the schema
- closes #33, addresses changes proposed there by including them in the new format used here

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,15 +9,12 @@ description = "LSL (Linden Scripting Language) definitions generator"
 readme = "README.md"
 license-files = ["LICENSE"]
 authors = [
-    {name = "Second Life", email = "opensource@lindenlab.com"},
-    {name= "Builder's Brewery", email = "buildersbrewery@gmail.com"}
+    {name = "Second Life", email = "opensource@lindenlab.com"}
 ]
 classifiers = [
     "Development Status :: 4 - Beta",
-
     "Intended Audience :: Developers",
     "Topic :: Software Development :: Build Tools",
-
     "Programming Language :: Python :: 3",
     "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",


buildersbrewery:
Anybody have a regex for argument names?

Does something like this work for LSL? 

Sublime Text 2025-12-03 21 54 58

lsl_definitions.schema.json
```diff
@@ -0,0 +1,467 @@
...
+                        "index-semantics": {
+                            "type": "boolean",
+                            "default": false
+                        },
+                        "linden-experience": {
+                            "type": "boolean",
+                            "default": false
+                        },
+                        "mono-sleep": {
+                            "$ref": "#definitions/nonNegativeNumber"
```
WolfGangS: Missing `/` after `#`
`#/definitions/nonNegativeNumber`

lsl_definitions.schema.json
```diff
@@ -0,0 +1,467 @@
...
+                                    "const": "quaternion"
+                                },
+                                { "const": "rotation" },
+                                { "const": "string" },
+                                { "const": "vector" },
+                                { "const": "void" }
+                            ]
+                        },
+                        "sleep": {
+                            "$ref": "#definitions/nonNegativeNumber"
```
WolfGangS: Missing `/` after `#`
`#/definitions/nonNegativeNumber`

lsl_definitions.schema.json
```diff
@@ -0,0 +1,467 @@
...
+                    },
+                    "required": [
+                        "tooltip"
+                    ]
+                }
+            },
+            "type": "object"
+        },
+        "llsd-lsl-syntax-version": {
+            "$ref": "#definitions/positiveInteger"
```
WolfGangS: Missing `/` after `#`
`#/definitions/positiveInteger`

lsl_definitions.schema.json
WolfGangS:
```suggestion
                            "$ref": "#/definitions/nonNegativeNumber"
```

lsl_definitions.schema.json
```diff
@@ -0,0 +1,467 @@
...
+                        }
+                    },
+                    "required": [
+                        "tooltip"
+                    ],
+                    "type": "object"
+                }
+            },
+            "type": "object"
+        },
```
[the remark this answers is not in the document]
buildersbrewery: Why are the values of the math constants in string format in this YAML file? Was there a technical reason? Can I just change them to use a float value in the YAML file and change the JSON schema for the YAML to use `number`?

image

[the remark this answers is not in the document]
WolfGangS: I suspect this file was originally generated from ll's llsd XML format. I might be wrong.

But if it was, pretty much everything in XML is a string till you use meta data to cast it. So possible it was just read as a string from the XML.

lsl_definitions.yaml
```diff
@@ -2164,15 +2168,15 @@ constants:
...
-    value: '3.14159265'
+    value: 3.14159265
   PING_PONG:
     tooltip: Play animation going forwards, then backwards.
     type: integer
     value: '0x8'
   PI_BY_TWO:
     tooltip: 1.57079633 - The number of radians in a quarter circle.
     type: float
-    value: '1.57079633'
```
HaroldCindy: Odd as it meant seem, floating point values are meant to be strings in this file. The values there are originally extracted from C++ files in the simulator, and we want to make sure the the interpretation of the floating point values doesn't change until the ultimate point-of-use.

lsl_definitions.yaml
```diff
@@ -3973,7 +3977,7 @@ constants:
   TWO_PI:
     tooltip: 6.28318530 - The radians of a circle.
     type: float
-    value: '6.28318530'
+    value: 6.28318530
```
HaroldCindy: see above about floating point

lsl_definitions.yaml
```diff
@@ -3653,7 +3657,7 @@ constants:
   SQRT2:
     tooltip: 1.41421356 - The square root of 2.
     type: float
-    value: '1.41421356'
+    value: 1.41421356
```
HaroldCindy: see above about floating point

buildersbrewery:
I have addressed all comments above. Haven't tested myself locally, yet. But everything is ready to go.

I'll switch to "ready for review" once I've tested locally, but feel free to do it yourself and provide feedback in the meantime.

pyproject.toml
```diff
@@ -24,20 +28,29 @@ classifiers = [
 requires-python = ">=3.8"
 dependencies = [
     "PyYAML",
-    "llsd",
-    "yamale",
+    "llsd"
+]
+keywords = ["Linden Scripting Language", "Second Life"]
+
+[project.optional-dependencies]
```
buildersbrewery: moved `jsonschema` to optional dependencies as requested

HaroldCindy:
Looks like CI is failing, think the command name in the workflow def needs updating

validate.py
```diff
@@ -0,0 +1,34 @@
...
+def validate_lsl_definitions_via_jsonschema() -> bool:
+    with open("lsl_definitions.schema.json", "r", encoding="utf-8") as schema_file:
+        schemadata = json.load(schema_file)
+    with open("lsl_definitions.yaml", "r", encoding="utf-8") as yaml_file:
+        yamldata = yaml.safe_load(yaml_file)
+
+    try:
+        jsonschema.validate(instance=yamldata, schema=schemadata)
+        print("\nSUCCESS: Valid against the schema")
+        return True
```
HaroldCindy: When this is used in utility script form, I believe the return of the function is treated as if it were an integer and used as the return code of the script. Would explain why CI is failing even with a valid schema.

HaroldCindy (approved):
LGTM, thanks!

PR #77, opened 2026-03-13
Recommend ll.ToLower in the string.lower docs. No Deprecation

tapple:
- Reversal of part of #74

### DIFF
--- a/generated/lua_keywords_pretty.xml
+++ b/generated/lua_keywords_pretty.xml
@@ -15056,6 +15056,8 @@ Returns true if result is non-zero.</string>
                   </map>
                </map>
             </array>
+            <key>deprecated</key>
+            <boolean>true</boolean>
             <key>energy</key>
             <real>10.0</real>
             <key>return</key>
@@ -15344,6 +15346,8 @@ Returns true if result is non-zero.</string>
                   </map>
                </map>
             </array>
+            <key>deprecated</key>
+            <boolean>true</boolean>
             <key>energy</key>
             <real>10.0</real>
             <key>return</key>
--- a/generated/lua_keywords_pretty.xml
+++ b/generated/lua_keywords_pretty.xml
@@ -14980,6 +14980,8 @@ Returns true if result is non-zero.</string>
                   </map>
                </map>
             </array>
+            <key>deprecated</key>
+            <boolean>true</boolean>
             <key>energy</key>
             <real>10.0</real>
             <key>return</key>
@@ -15268,6 +15270,8 @@ Returns true if result is non-zero.</string>
                   </map>
                </map>
             </array>
+            <key>deprecated</key>
+            <boolean>true</boolean>
             <key>energy</key>
             <real>10.0</real>
             <key>return</key>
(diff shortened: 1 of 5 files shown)


tapple:
I added this to the d.luau file:
```lua
declare extern type string with
  @[deprecated {use='ll.ToLower', reason='Unicode support.'}] function lower(self): string
end
```
It seems to work with variables, but not literals, and causes other problems with literals in strict mode:
```lua
--!strict
local _ = string.lower("HELLO")
local s: string = "HELLO"
local _ = tostring(NULL_KEY):lower()
local _ = URL_REQUEST_DENIED:lower()
local _ = s:lower()
```

image

```
$ ../../secondlife/luau-lsp/build/luau-lsp analyze --defs sl-vscode-plugin/slua_fixes.d.luau string_test.luau 
[INFO] Loading definitions file: @roblox - sl-vscode-plugin/slua_fixes.d.luau
[INFO] sourcemap is disabled - automatic updating on sourcemap changes disabled
string_test.luau(2,24): TypeError: Expected this to be 'string', but got 'string'
string_test.luau(3,1): TypeError: Expected this to be 'string', but got 'string'
string_test.luau(2,11): DeprecatedApi: Member 'string.lower' is deprecated, use 'll.ToLower' instead. Unicode support.
string_test.luau(5,11): DeprecatedApi: Member 'URL_REQUEST_DENIED.lower' is deprecated, use 'll.ToLower' instead. Unicode support.
string_test.luau(6,11): DeprecatedApi: Member 'lower' is deprecated, use 'll.ToLower' instead. Unicode support.
```

Issues:
1. False error `Expected this to be 'string', but got 'string'` on lines 2 and 3 (only with `--!strict`)
2. Missing `DeprecatedApi` error on lines 3 and 4

This works for regular extern types, just not string:
```lua
declare extern type uuid with
  @[deprecated {use="logic", reason="reason"}] function randomize(self): uuid
end
```

image


string seems to be magic, which isn't terribly surprising

NikkiLacrima:
I think that removing functions from standard lua libraries is a BAD idea. Let the users choose unless very bad things happens. Being able to do the same thing in two ways is not IMHO a problem, its a feature.

Martin-Pitt:
A removal is very different from deprecation, that's not what is being done here.

Deprecation is fundamentally only used to discourage the usage of an API that it marks, nothing more. This _could_ be because it is not useful, superseded or earmarked for removal.

In this case the reason here is that it might not be doing its job properly when an alternative that does exists. There are cases where having fast cheap ASCII-only functions is actually, quite, useful. Lowering case when a string could have international characters, however not so much.

HaroldCindy:
Hmmm. I'm iffy on this one if it's difficult to apply across the board, especially if there's genuine usecases for `string.lower()` vs `ll.ToLower()`.  if you only want _ASCII_ lowering. It's a little complicated since `string` in Luau does double-duty as both the `bytes` and `string` implementation, and I've intentionally used `bytes.lower()` in Python.

Perhaps a caveat in the description calling this out would suffice for now?

tapple:
Changed the PR from deprecating to just documenting the difference: https://github.com/secondlife/lsl-definitions/pull/77/changes/636ab699cd35aa24db1c677599d1b36b49bbbe84

HaroldCindy (approved):

PR #84, opened 2026-03-25
Deprecate `ll.Frand` for `math.random`

tapple:
- Extracted from #71 for further discussion

`math.random` has two modes: float mode (0 arguments):
```
> math.random()
0.5956555082509403
> math.random()
0.526518904998895
> math.random()
0.8429784570452903
> math.random()
0.49559577154027495
```

and integer mode (1-2 arguments):
```
> math.random(5, 10)
8
> math.random(5, 10)
9
> math.random(-5, 10)
10
> math.random(-5, 10)
-4
> math.random(10)
1
> math.random(10)
1
```

`math.random() * x` is a direct replacement for `ll.Frand(x)`, even though `math.random(x)` is not.

`math.random(x)` is a direct replacement for `math.modf(ll.Frand(x))` (`(integer)llFrand(x)` in lsl) though, which, at least in my own code, is far more common.

### DIFF
--- a/generated/lua_keywords_pretty.xml
+++ b/generated/lua_keywords_pretty.xml
@@ -19019,6 +19019,8 @@ Returns true if result is non-zero.</string>
                   </map>
                </map>
             </array>
+            <key>deprecated</key>
+            <boolean>true</boolean>
             <key>energy</key>
             <real>10.0</real>
             <key>return</key>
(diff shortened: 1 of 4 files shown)


HaroldCindy:
Honestly, I think we're going to punt on this one. Depending on your usecase it might be more reasonable / natural to reach for one or the other, and both functions have their own pitfalls at the moment. The RNG interface in general will require some deeper thought in the future :(

PR #104, opened 2026-04-13
Classified all the LSL constants into enums

tapple:
- Part of #23 
- Part of #31
- Related: https://github.com/secondlife/create

Adds an `enums` section to `lsl_definitions.yaml`, for the purpose of documentation generation efforts, such as https://create.secondlife.com. And classify (almost) all of LSL's integer constants into one or more of these enums:
```
enums:
# This section documents the relationship between LSL's functions and constants.
# Neither LSL nor Lua have real enums, nor is this a proposal to add them.
# This is purely for the purpose of documentation generation, and to
# document these relationships somewhere other than the LSL wiki.
```
The function -> enum side of the relationship will be done in a follow-up PR

PR #105, opened 2026-04-16
Deprecate obsolete constants

tapple:
- Extracted from #104

Deprecate these two constants:
- `PRIM_MATERIAL_LIGHT`: use `PRIM_FULLBRIGHT`
- `STATUS_CAST_SHADOWS`: not implemented
Also fixed a comment about `CLICK_ACTION_TOUCH`


lsl_definitions.yaml
```diff
@@ -1218,6 +1217,8 @@ constants:
     value: 9
   CLICK_ACTION_NONE:
     member-of: [ClickAction]
+    deprecated:
+      use: CLICK_ACTION_TOUCH
```
tapple: - Revert this one, correct? https://github.com/secondlife/lsl-definitions/pull/104#discussion_r3091130330

lsl_definitions.yaml
```diff
@@ -2144,6 +2145,8 @@ constants:
     value: 2
   LAND_LARGE_BRUSH:
     member-of: [ModifyLandBrush]
+    deprecated:
+      use: '2'
```
tapple: - Alternate proposal: #106

lsl_definitions.yaml
tapple: Reverted

HaroldCindy (approved):

PR #107, opened 2026-04-18
Note llGetPrimitiveParams([PRIM_PROJECTOR]) bug

tapple:
just adding more info to a comment

PR #108, opened 2026-04-19
Generate a fluent builder interface for llSetPrimitiveParams

HaroldCindy:
Uses the enums and rulesets we already have under the hood to generate a builder definition and its typings whenever we add a new rule.

Example:

```lua
-- Simple use, apply params immediately
llprim.ParamsSetter.new()
  :color(ALL_SIDES, vector(0,0,1), 1)
  :text("first", vector.one, 1)
  :apply(LINK_ROOT) -- Uses LINK_THIS if a number not passed


-- But you can also collect rules to build upon and apply later
-- Here we arrange the child links in a line above the root
local rules = llprim.ParamsSetter.new()

for i=2,ll.GetNumberOfPrims() do
  -- The method calls really just append an SPP rule then return `self`
  rules
    :targetLink(i)
    :posLocal(vector(0,0,1 * i))
end

-- And we apply the changes atomically as one `ll.SetLinkPrimitiveParamsFast()` under the hood.
rules:apply()
```


lsl_definitions/generators/keywords.py
```diff
@@ -101,7 +101,7 @@ def dump_slua_syntax(
             syntax["functions"].update(module.to_keywords_functions_dict())
     syntax["functions"].update(ll_module.to_keywords_functions_dict())
     for func in sorted(definitions.functions.values(), key=lambda x: x.name):
-        if not func.private:
+        if not func.private and not func.slua_removed:
```
HaroldCindy, line 104: Submarine change, these were getting shifted around and they're definitely not in `ll.` so I omitted them.

lsl_definitions/generators/slua.py
```diff
@@ -101,12 +99,23 @@ def gen_selene_yml(definitions: LSLDefinitions, slua_definitions: SLuaDefinition
     def selene_type(type_str: str, default="any") -> str | dict | None:
         if type_str.endswith("?"):
             type_str = type_str[:-1]
+        # GLTF-clearable params are encoded as ``<base> | ""``. Selene args can't
+        # be unions, so collapse to the base type if selene accepts a bare string
+        # there; otherwise emit a display-only hint mentioning both forms.
+        clear_sentinel = ' | ""'
```
HaroldCindy, line 106: I'm not entirely certain if this is a common idiom elsewhere but seemed worth some trivial support for it rather than defaulting to `any`. Might need another for `| false` and `| true`.

lsl_definitions/generators/slua.py
```diff
@@ -101,12 +99,23 @@ def gen_selene_yml(definitions: LSLDefinitions, slua_definitions: SLuaDefinition
...
+            if base == "string":
+                return "string"
+            return {"display": f"{base_str} | string"}
         type_map = {
             "boolean": "bool",
             "boolean | number": "bool",
             "number": "number",
             "string": "string",
             "string | uuid": "string",
+            "(string | uuid)": "string",
```
HaroldCindy, line 116: I liked the parens and it _feels_ more correct to me even if `string | uuid?` is semantically the same.

lsl_definitions/lsl.py
```diff
@@ -488,7 +488,7 @@ class LSLDefinitions(NamedTuple):
     constants: Dict[str, LSLConstant]
     controls: dict
     types: dict
-    rulesets: dict
+    builder_rulesets: dict
```
HaroldCindy, line 491: More honest about the fact that this is a special kind of ruleset that probably won't be used elsewhere.

lsl_definitions.schema.json
```diff
@@ -447,55 +447,72 @@
...
             "type": "object"
         },
-        "rulesets": {
+        "builder-rulesets": {
             "additionalProperties": false,
+            "markdownDescription": "Rulesets that codegen into fluent-builder SLua classes (ordered rule+value setters, face-target, discriminator variants).",
             "type": "object",
             "properties": {
-                "prim_params": {
+                "prim-params": {
```
HaroldCindy, line 455: I did a general kebab-ification since apparently I forgot that style point in my first pass.

generated/secondlife.d.luau
```diff
@@ -169,6 +169,66 @@ declare extern type LLTimers with
   function off(self, callback: LLTimerCallback): boolean
 end
 
+export type ParamsSetterTypeMeta = {
+  __index: ParamsSetterTypeMeta,
+  targetLink: <T>(self: T & ParamsSetterType, link_target: number) -> T,
```
HaroldCindy: The `T & ParamsSetterType` pattern may seem curious. The reason I went with that approach is that someone may actually use a metatable that has _additional_ methods, and I didn't want to lose that type information in the `return self`. This way, we just check that `T` satisfies `ParamsSetterType` and return `T` with the correct type info.

lsl_definitions/rulesets.py
```diff
@@ -0,0 +1,128 @@
...
+@dataclasses.dataclass(frozen=True)
+class BuilderMethod:
+    # `face_target` and `nullable` are SPP-flavored
+    name: str
+    """Final, already-resolved method name on the builder class."""
+    rule_name: str
+    """Owning ruleset rule, prefix stripped (e.g. `COLOR` or `TYPE`)"""
+    rule_tag: int
+    """Int value of `rule_name`"""
+    variant_tag: Optional[int]
```
HaroldCindy, line 53: These aren't used by the code in this repo, but could be used by codegen that generates the supporting code for these methods

generated/lua_keywords_pretty.xml
```diff
@@ -19218,22 +19237,6 @@ Returns true if result is non-zero.</string>
             <key>tooltip</key>
             <string>Returns the alpha value of Face.\nReturns the 'alpha' of the given face. If face is ALL_SIDES the value returned is the mean average of all faces.</string>
          </map>
-         <key>ll.GetAndResetTime</key>
```
tapple, line 19221: Whoops. nice catch

generated/secondlife.d.luau
```diff
@@ -169,6 +169,66 @@ declare extern type LLTimers with
...
+  physicsMaterial: <T>(self: T & ParamsSetterType, flag: number) -> T,
+  physics: <T>(self: T & ParamsSetterType, enabled: boolean) -> T,
+  tempOnRez: <T>(self: T & ParamsSetterType, enabled: boolean) -> T,
```
tapple: temporary is probably a better name. TEMP_ON_REZ hasn't been an accurate name for about 20 years now

generated/secondlife.d.luau
```diff
@@ -169,6 +169,66 @@ declare extern type LLTimers with
   function off(self, callback: LLTimerCallback): boolean
 end
 
+export type ParamsSetterTypeMeta = {
+  __index: ParamsSetterTypeMeta,
+  targetLink: <T>(self: T & ParamsSetterType, link_target: number) -> T,
+  physicsMaterial: <T>(self: T & ParamsSetterType, flag: number) -> T,
+  physics: <T>(self: T & ParamsSetterType, enabled: boolean) -> T,
```
tapple: `physical` would be a better name

generated/secondlife.d.luau
```diff
@@ -169,6 +169,66 @@ declare extern type LLTimers with
...
+  primTypePrism: <T>(self: T & ParamsSetterType, hole_type: number, cut: vector, hollow: number, twist: vector, top_size: vector, top_shear: vector) -> T,
+  primTypeSphere: <T>(self: T & ParamsSetterType, hole_type: number, cut: vector, hollow: number, twist: vector, dimple: vector) -> T,
+  primTypeTorus: <T>(self: T & ParamsSetterType, hole_type: number, cut: vector, hollow: number, twist: vector, hole_size: vector, top_shear: vector, advanced_cut: vector, taper: vector, revolutions: number, radius_offset: number, skew: number) -> T,
+  primTypeTube: <T>(self: T & ParamsSetterType, hole_type: number, cut: vector, hollow: number, twist: vector, hole_size: vector, top_shear: vector, advanced_cut: vector, taper: vector, revolutions: number, radius_offset: number, skew: number) -> T,
+  primTypeRing: <T>(self: T & ParamsSetterType, hole_type: number, cut: vector, hollow: number, twist: vector, hole_size: vector, top_shear: vector, advanced_cut: vector, taper: vector, revolutions: number, radius_offset: number, skew: number) -> T,
+  primTypeSculpt: <T>(self: T & ParamsSetterType, texture: (string | uuid), sculpt_type: number) -> T,
```
tapple: should these maybe be named `typeRing` and `typeBox` instead?

generated/secondlife.d.luau
```diff
@@ -296,6 +356,15 @@ declare lljson: {
 }
 
 
+---------------------------
+-- Global Table: llprim
+---------------------------
+
+declare llprim: {
```
tapple, line 363: will there be one of these global tables per wrapped function? not one table for all of them? dunno if you've more planned

lsl_definitions/generators/slua.py
tapple: I only remember gltf being like this

I still haven't figured out what selene actually uses its type field for. I can't ever get it to show type errors

I'd rather it just showed whatever <base> would turn into below. so, just  replace this whole if block with
```
type_str.removesuffix(clear_sentinel)
```

lsl_definitions/rulesets.py
```diff
@@ -0,0 +1,128 @@
...
+    # LSL-side semantic validation already ran in
+    # `LSLDefinitionParser._validate_builder_rule_variants`.
+    ruleset = lsl.builder_rulesets["prim-params"]
+    rule_enum = lsl.enums[ruleset["enum"]]
+
+    methods: List[BuilderMethod] = []
+    for rule_name, rule in ruleset["rules"].items():
+        face_target = bool(rule.get("face-target"))
+        nullable = bool(rule.get("nullable"))
+        rule_tag = rule_enum.by_name[rule_name].value
```
tapple, line 94: in the rest of the code, these would have already been loaded into a dataclass with `bool` `bool` `enum` fields. Why not here? still a prototype subject to change? too specific to this one construct?

generated/secondlife.d.luau
HaroldCindy: The "prim" prefix is intentional since "type" is ambiguous. It's specifically making this a prim of type <x>.

generated/secondlife.d.luau
HaroldCindy: No but I imagine `ll*` namespaces will continue to be added as reasonable / necessary. This didn't clearly belong anywhere else, so an `llprim` seemed like the logical place to put it.

Interested in other approaches though. I originally had it in `llwrappers` but it seemed more proper to categorize it by its intended use rather than how it's implemented.

generated/secondlife.d.luau
```diff
@@ -169,6 +169,66 @@ declare extern type LLTimers with
   function off(self, callback: LLTimerCallback): boolean
 end
 
+export type ParamsSetterTypeMeta = {
```
tapple: should this be `PrimitiveParamsSetterType`? it's not reusable for any other type of params list

tapple (approved):

generated/secondlife.d.luau
tapple: oh is that the incantation to get that working. I'm having this issue in my own code

generated/secondlife.d.luau
```diff
@@ -169,6 +169,66 @@ declare extern type LLTimers with
...
+  primTypeRing: <T>(self: T & ParamsSetterType, hole_type: number, cut: vector, hollow: number, twist: vector, hole_size: vector, top_shear: vector, advanced_cut: vector, taper: vector, revolutions: number, radius_offset: number, skew: number) -> T,
+  primTypeSculpt: <T>(self: T & ParamsSetterType, texture: (string | uuid), sculpt_type: number) -> T,
+  texture: <T>(self: T & ParamsSetterType, face: number, texture: (string | uuid), repeats: vector, offsets: vector, rotation_in_radians: number) -> T,
+  color: <T>(self: T & ParamsSetterType, face: number, color: vector, alpha: number) -> T,
+  bumpShiny: <T>(self: T & ParamsSetterType, face: number, shiny: number, bump: number) -> T,
+  fullbright: <T>(self: T & ParamsSetterType, face: number, enabled: boolean) -> T,
+  flexible: <T>(self: T & ParamsSetterType, enabled: boolean, softness: number, gravity: number, friction: number, wind: number, tension: number, force: vector) -> T,
+  texgen: <T>(self: T & ParamsSetterType, face: number, mode: number) -> T,
```
HaroldCindy: Hmm. Realistically, maybe this is best exposed as `[boolean, planar_mapping]` since there's only default and planar mapping modes.

generated/secondlife.d.luau
```diff
@@ -169,6 +169,66 @@ declare extern type LLTimers with
...
+  primTypeTorus: <T>(self: T & ParamsSetterType, hole_type: number, cut: vector, hollow: number, twist: vector, hole_size: vector, top_shear: vector, advanced_cut: vector, taper: vector, revolutions: number, radius_offset: number, skew: number) -> T,
+  primTypeTube: <T>(self: T & ParamsSetterType, hole_type: number, cut: vector, hollow: number, twist: vector, hole_size: vector, top_shear: vector, advanced_cut: vector, taper: vector, revolutions: number, radius_offset: number, skew: number) -> T,
+  primTypeRing: <T>(self: T & ParamsSetterType, hole_type: number, cut: vector, hollow: number, twist: vector, hole_size: vector, top_shear: vector, advanced_cut: vector, taper: vector, revolutions: number, radius_offset: number, skew: number) -> T,
+  primTypeSculpt: <T>(self: T & ParamsSetterType, texture: (string | uuid), sculpt_type: number) -> T,
+  texture: <T>(self: T & ParamsSetterType, face: number, texture: (string | uuid), repeats: vector, offsets: vector, rotation_in_radians: number) -> T,
+  color: <T>(self: T & ParamsSetterType, face: number, color: vector, alpha: number) -> T,
+  bumpShiny: <T>(self: T & ParamsSetterType, face: number, shiny: number, bump: number) -> T,
```
HaroldCindy: I just now realized that `LSL_BUMP_SHINY` takes the params in the order `[shiny, bump]`... Maybe we should correct that by renaming the method to `shinyBump`?

generated/secondlife.d.luau
```diff
@@ -169,6 +169,66 @@ declare extern type LLTimers with
...
+  gltfMetallicRoughness: <T>(self: T & ParamsSetterType, face: number, texture: (string | uuid) | "", repeats: vector | "", offsets: vector | "", rotation_in_radians: number | "", metallic_factor: number | "", roughness_factor: number | "") -> T,
+  gltfBaseColor: <T>(self: T & ParamsSetterType, face: number, texture: (string | uuid) | "", repeats: vector | "", offsets: vector | "", rotation_in_radians: number | "", linear_color: vector | "", alpha: number | "", gltf_alpha_mode: number | "", alpha_mask_cutoff: number | "", double_sided: boolean | "") -> T,
+  renderMaterial: <T>(self: T & ParamsSetterType, face: number, render_material: (string | uuid)) -> T,
+  sitFlags: <T>(self: T & ParamsSetterType, flags: number) -> T,
+  damage: <T>(self: T & ParamsSetterType, damage: number, damage_type: number) -> T,
+  health: <T>(self: T & ParamsSetterType, health: number) -> T,
+  collisionSound: <T>(self: T & ParamsSetterType, sound: (string | uuid), volume: number) -> T,
+  apply: (self: ParamsSetterType, link: number?) -> (),
```
HaroldCindy: if not passed, targets `LINK_THIS`, so similar to `llSetPrimitiveParams()`.

lsl_definitions/rulesets.py
HaroldCindy: Yeah, it's basically specific to this one construct so I didn't want to try to generalize it too much.

generated/secondlife.d.luau
tapple: I wondered the same thing about https://wiki.secondlife.com/wiki/LlScriptProfiler . Though, I think llScriptProfiler is more likely than PRIM_TEXGEN to get new modes soon

Rider-Linden (approved):

tapple (approved):

tapple:
- part of https://github.com/secondlife/slua/issues/70

lsl_definitions/generators/slua.py
HaroldCindy: Aha, can't use `removesuffix` because this needs to run under Python 3.7, but will do the equivalent :)

lsl_definitions.yaml
```diff
@@ -14480,67 +14493,66 @@ rulesets:
...
         params:
-          # This should almost certainly have asset semantics, but the server
-          # appears to only allow `string`, though stringified UUID is fine?
-          # Will need to fix that, but we need `asset` here since the `get`
-          # may still return a UUID.
-          - [ asset, texture ]
+          # TODO: This should almost certainly have asset semantics, but the server
+          #  appears to only allow `string`, though stringified UUID is fine?
+          #  Will need to fix that.
+          - [ string, texture ]
```
tapple: - I left further comments about this in #107

generated/secondlife.d.luau
tapple: yeah. as long as there don't wind up being dozens. I already think llbase64 is a bit too specialized. maybe llcodecs or llmime instead. that would also leave room for encryption, urlencode, Though, urlencode could also logically be in llhttp, which I'm sure is planned

But yeah, llprim can logically be home to quite a few more functions. it's all right

tapple: also, by this logic, `LLEvents` and `LLTimers` should probably be in a table. But, `LLEvents` is transitional, like `ll`, so it's exempt.

But `LLTimers` is new. maybe it should be in a table. perhaps `llscript`, along with other things related to scripts and script execution, like [metadata](https://feedback.secondlife.com/slua-alpha/p/script-meta-information), [memory usage](https://feedback.secondlife.com/slua-alpha/p/ability-to-get-a-variables-memory-usage-slua-only-maybe), [event wrappers](https://feedback.secondlife.com/slua-alpha/p/custom-event-error-handler).

PR #110, opened 2026-04-20
SLua - Deprecate `ll.GetUnixTime` in favor of os.time

WolfGangS:
No reason to encourage use of lsl's method when there is a more flexible native version.


WolfGangS:
Closing in favor of #111

PR #111, opened 2026-04-20
Deprecate `ll.GetUnixTime` in favor of `os.time`

tapple:
- Better version of #110 

Deprecate ll.GetUnixTime and document other lsl equivilencies:
- `ll.GetUnixTime`: deprecated in favor of `os.time` due to integer overflow
- `ll.GetDate`: documented the equivilancy with `os.date`
- `ll.GetTimeStamp`: documented the near-equivilancy with `os.date`
- `ll.GetTime` and `os.clock`: documented their slightly different behaviors

### DIFF
--- a/generated/secondlife.d.luau
+++ b/generated/secondlife.d.luau
@@ -769,7 +769,7 @@ declare ll: {
   GetTimeOfDay: () -> number,
   GetTimestamp: () -> string,
   GetTorque: () -> vector,
-  GetUnixTime: () -> number,
+  GetUnixTime: @[deprecated {use='os.time', reason='Int32 will wrap on 2038-01-19'}]() -> number,
   GetUsedMemory: () -> number,
   GetUsername: (AvatarID: uuid) -> string,
   GetVel: () -> vector,
--- a/generated/lsl_keywords_pretty.xml
+++ b/generated/lsl_keywords_pretty.xml
@@ -15221,7 +15221,7 @@ If another state is defined before the default state, the compiler will report a
             <key>sleep</key>
             <real>0.0</real>
             <key>tooltip</key>
-            <string>Returns the time in seconds since the last region reset, script reset, or call to either llResetTime or llGetAndResetTime.</string>
+            <string>Returns the time in seconds since the last region reset, script reset, or call to either llResetTime or llGetAndResetTime. Has a resolution of 0.022s (1 server frame), and is 6-11x faster to look up than lua's os.clock</string>
          </map>
          <key>llGetTimeOfDay</key>
          <map>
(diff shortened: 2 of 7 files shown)


generated/secondlife.d.luau
tapple, line 768: what the heck. how is `ll.GetTime` 11x faster than `os.clock`:

```lua
for i = 1, 4400 do local _ = ll.GetTime() end
print(os.clock(), ll.GetUsedMemory() - 2302) -- 0.00277799998507835    57
```

```lua
for i = 1, 400 do local _ = os.clock() end
print(os.clock(), ll.GetUsedMemory() - 2302) -- 0.0027720000023487954    42
```

`os.clock` seems to have 17 more bits of precision:
```lua
for i = 1, math.random(1_000_000) do end
local t = ll.GetTime()
local c = os.clock()
local tf, te = math.frexp(t)
local cf, ce = math.frexp(c)
print("ll.GetTime", tf * 2^53, te, tf * 2^36) --  ll.GetTime    6306381550977024    1    48113872917
print("os.clock", cf * 2^53, ce) -- os.clock    6306831910939761    1
```
`os.clock` has full `float64` (53-bit) precision. `ll.GetTime` has 36-bit precision, somewhere between a `float32` (24-bit) and `float64` (53-bit). Though the addid precision below `2^-36=1.5e-11` is in the picosecond range, and thus pure noise. I believe both functions are truncated to milisecond precision. I'm unsure how this truncation is done. Not in a way that makes the low bits/decimals 0; that's for sure

tapple: Better test script:
```lua
local lastTime = -1
local iterCount = 0
local ticks = {}
while lastTime < 0.01 do
    local t = ll.GetTime()
--    local t = os.clock()
    iterCount += 1
    if t ~= lastTime then
        table.insert(ticks, t)
        lastTime = t
    end
end

print("iterCount", iterCount, "ticks", lljson.slencode(ticks))
```

With `ll.GetTime()`:
> [11:11 AM] Object: iterCount    3747    ticks    [0.0025739999837242,0.021947000001092]


With `os.clock()`:
> [11:10 AM] Object: iterCount    659    ticks    [0.0029199999989942,0.0030199999989942,0.0031199999989942,0.022477999969851]

`ll.GetTime` seems to update only once per server frame, and is thus much faster to compute (probably just a variable lookup internally). It thus has a resolution of `1/45 = 0.022`s.

`os.clock()` seems attempt to get an accurate time every call, and is thus slower. It appears to have a resolution of `0.0001`s. This is only 1/3 the length of a script time slice, so `os.clock()` can really only return 1 of 3 values during a frame.

HaroldCindy (approved):
LGTM, thanks!

PR #113, opened 2026-04-27
Associate LSL Functions with Enums

tapple:
- Follow up to #104 
- Part of #31 
- Closes #23

Add new annotations to function arguments:
- enum-semantics
- param-semantics
- param-get-semantics

Add new annotations to functions:
- enum-semantics
- param-get-semantics

Add new enum types:
- param-list
- paam-dict

PR #114, opened 2026-04-27
Add missing constant `SIT_OK`

tapple:
According to [the wiki](https://wiki.secondlife.com/wiki/LlSitOnLink), `llSitOnLink` has 8 possible return values. But, only 7 of them are named constants. Add the missing one.

Discovered while auditing the `Error` enums for #113

### DIFF
--- a/generated/cpp/lllslconstants_generated.h
+++ b/generated/cpp/lllslconstants_generated.h
@@ -841,6 +841,7 @@ constexpr int32_t LSL_GEN_SIT_NOT_EXPERIENCE = -1;
 constexpr int32_t LSL_GEN_SIT_NO_ACCESS = -6;
 constexpr int32_t LSL_GEN_SIT_NO_EXPERIENCE_PERMISSION = -2;
 constexpr int32_t LSL_GEN_SIT_NO_SIT_TARGET = -3;
+constexpr int32_t LSL_GEN_SIT_OK = 1;
 constexpr int32_t LSL_GEN_SKY_ABSORPTION_CONFIG = 16;
 constexpr int32_t LSL_GEN_SKY_AMBIENT = 0;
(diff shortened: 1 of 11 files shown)


Martin-Pitt:
If successful returns 1 you mean?

tapple:
yes. I defined SIT_OK = 1

generated/cpp/lllslconstants_generated.h
```diff
@@ -841,6 +841,7 @@ constexpr int32_t LSL_GEN_SIT_NOT_EXPERIENCE = -1;
 constexpr int32_t LSL_GEN_SIT_NO_ACCESS = -6;
 constexpr int32_t LSL_GEN_SIT_NO_EXPERIENCE_PERMISSION = -2;
 constexpr int32_t LSL_GEN_SIT_NO_SIT_TARGET = -3;
+constexpr int32_t LSL_GEN_SIT_OK = 1;
```
tapple, line 844: The wiki is correct: this script seats me and prints `1`:
```lsl
default { state_entry() {
llSitTarget(<0, 0, 1>, ZERO_ROTATION);
llOwnerSay((string)llSitOnLink(llGetOwner(), LINK_THIS));
}}
```
> [12:13 PM] Object: 1

lsl_definitions.yaml
```diff
@@ -4932,6 +4932,11 @@ constants:
     tooltip: No available sit target in linkset for forced sit.
     type: integer
     value: -3
+  SIT_OK:
+    member-of: [SitError]
```
HaroldCindy, line 4936: 🔕 May have to do another pass at these names later, "error" no longer fits. Doesn't block this at all.

HaroldCindy (approved):

PR #115, opened 2026-04-27
Add missing constant `ENV_OK`

tapple:
According to the wiki, the environment modification functions have 10 possible return values. But, only 9 of them are named constants. Add the missing one.

- https://wiki.secondlife.com/wiki/LlReplaceAgentEnvironment
- https://wiki.secondlife.com/wiki/LlSetAgentEnvironment
- https://wiki.secondlife.com/wiki/LlReplaceEnvironment
- https://wiki.secondlife.com/wiki/LlSetEnvironment


generated/cpp/lllslconstants_generated.h
```diff
@@ -209,6 +209,7 @@ constexpr int32_t LSL_GEN_ENV_NO_ENVIRONMENT = -3;
 constexpr int32_t LSL_GEN_ENV_NO_EXPERIENCE_LAND = -7;
 constexpr int32_t LSL_GEN_ENV_NO_EXPERIENCE_PERMISSION = -2;
 constexpr int32_t LSL_GEN_ENV_NO_PERMISSIONS = -9;
+constexpr int32_t LSL_GEN_ENV_OK = 1;
```
tapple, line 212: The wiki is correct; I tested all 4 functions, and they all 4 changed my environment and returned 1:
```lsl
default { state_entry() {
//llOwnerSay((string)llReplaceEnvironment(llGetPos(), "Orac - Drawing blue", -1, -1, -1));
//llOwnerSay((string)llSetEnvironment(llGetPos(), [SKY_AMBIENT, <0, 1, 0>]));
//llOwnerSay((string)llReplaceAgentEnvironment(llGetOwner(), 1, "Orac - Drawing red"));
//llOwnerSay((string)llSetAgentEnvironment(llGetOwner(), 1, [SKY_AMBIENT, <0, 1, 0>]));
}}
```
environment assets from the library

HaroldCindy:
@Rider-Linden Thoughts on this convention? I think it seems reasonable, since both the error case and the success case use non-zero return values, so `if (retval)` is never sensible. Currently people have to do `== 1` if they want to specifically check for success, and this would give a better constant for it.

While `TRUE` is incidentally also `1`, this isn't really a function that returns a boolean semantically, so `== TRUE` would be a little weird, and people might think that it could be made easier to read by changing to `if (retval)`, which would _not_ do the same thing.

tapple:
`if ll.ReplaceEnvironment(ll.GetPos(), "Orac - Drawing blue", -1, -1, -1) == TRUE` is invalid in lua, unless you define your own `TRUE` constant

HaroldCindy:
@tapple Indeed, that was strictly about this constant's applicability to LSL as well

Rider-Linden (approved):
Yeah.  This is entirely a good thing.  Not sure why I didn't add it at the time.

PR #119, opened 2026-05-07
Move parts of PrimParamsSetterType to yaml

tapple:
Move the non-generated parts of `PrimParamsSetterType` to `slua_definitions.yaml` so they can be documented better.

I also added another field to classes: `functions`, for static methods and constructors and such. They are rendered almost identically to methods, with a slight difference for selene. But mostly, I figure we'll have more of those in the future and a document generator probably wants to put them in different sections. The first class function is `PrimParamsSetterType.new`

I thought about moving `LLEvents.touch_start` and such from properties to functions, as they would fit. But, I kept them as properties since it doesn't make much sense to ever call them, only set them yourself

### DIFF
--- a/lsl_definitions/slua.py
+++ b/lsl_definitions/slua.py
@@ -259,7 +259,7 @@ class SLuaModule:
     """Module declaration with constants and functions"""
 
     name: str
-    callable: Optional[SLuaFunction] = None
+    callable: Optional[SLuaFunction]
     constants: List[SLuaProperty]
     functions: List[SLuaFunction]
     comment: str = ""
--- a/lsl_definitions/slua.py
+++ b/lsl_definitions/slua.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import abc
 import dataclasses
 import re
-from typing import TYPE_CHECKING, Any, List, Set, TextIO
+from typing import TYPE_CHECKING, Any, List, Optional, Set, TextIO
 
 import llsd
 import yaml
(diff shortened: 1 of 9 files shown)


lsl_definitions/slua.py
```diff
@@ -254,7 +259,7 @@ class SLuaModule:
     """Module declaration with constants and functions"""
 
     name: str
-    callable: Optional[SLuaFunction]
+    callable: SLuaFunction | None
```
HaroldCindy: IIRC `Optional` is necessary because Python 3.7

HaroldCindy (changes requested):
LGTM, the removed `Optional`s just need adding back in

lsl_definitions/slua.py
tapple: there was both `Optional` and `| None` in the same class before. checking if it runs on python 3.7

tapple: it runs on python 3.7. I can restore `Optional` though

lsl_definitions/slua.py
tapple, line 61: See. there was a `bool | None` here already, and it's been working. I just made it consistent

lsl_definitions/slua.py
tapple: - reverted and moved to #121

lsl_definitions/slua.py
tapple: - reverted and moved to #121

HaroldCindy (approved):

PR #120, opened 2026-05-08
Textmate Syntax file generation for LSL and SLua

### LINKED ISSUE #9: Add TextMate Grammar for Linden Scripting Language (LSL)
We’re looking for a high-quality TextMate grammar for Linden Scripting Language (LSL) to be added to this repository so that it can ultimately be be contributed to [tm-grammars](https://github.com/shikijs/textmate-grammars-themes/tree/main/packages/tm-grammars), the canonical grammar collection used by [Shiki](https://shiki.style/).

##  Why This Matters

We are [looking to](https://github.com/secondlife/issues/issues/1) use [Mintlify](https://mintlify.com/) for documentation, which relies on Shiki for syntax highlighting. LSL code blocks are currently unhighlighted, making developer docs harder to read and write. Fixing this will benefit the Second Life scripting community and enhance the developer experience across the web wherever Shiki is used.

## Acceptance Criteria

- Pull request opened against [lsl-definitions](https://github.com/secondlife/lsl-definitions)
- A fully functional .tmLanguage.json grammar for LSL
- Proper classification of LSL keywords, types, constants, events, functions, and operators
- Verified syntax highlighting using the Shiki engine (e.g., via shiki-playground)
- LSL added to the languages.json mapping if necessary
- Demo screenshot or GIF showing LSL highlighting in action

WolfGangS:
Adds basic template base generation of textmate syntax files for both LSL and JSON

I chose to use template that I insert the relevant parts into instead of generating the whole document in a schema, as that would be ALLOT more code, unless more libraries were introduced and i am under the impression less dependencies and so forth is better, due to places this may need to be run.

Closes #9


gen_all_definitions.sh
tapple: need a `mkdir syntax`

WolfGangS:
~~Known missing~~

slua_definitions.schema.json
```diff
@@ -317,6 +317,25 @@
                 }
             },
             "type": "object"
+        },
+        "metamethods": {
+            "items": { "$ref": "#/$defs/function" },
```
tapple: delete

slua_definitions.schema.json
```diff
@@ -317,6 +317,25 @@
...
+            "additionalProperties": false,
+            "patternProperties": {
+                "^[a-z]*": {
```
tapple:
```suggestion
                "^__[a-z]*": {
```

slua_definitions.schema.json
```diff
@@ -317,6 +317,25 @@
...
+                    "properties": {
+                        "tooltip": {
+                            "markdownDescription": "A brief description of this type.",
```
tapple:
```suggestion
                            "markdownDescription": "A brief description of this metamethod.",
```

slua_definitions.yaml
```diff
@@ -2812,3 +2812,48 @@ builtin-functions:
...
+  __unm:
+    tooltip: "Overload method for - as unaryoperator."
+  __iter:
+    tooltip: "Method called to iterate over the table."
+  __idiv:
+    tooltip: "Overload method for // operator."
+  __tojson:
+    tooltip: "Called by lljson encode and slencode to encode the value as JSON."
+  __jsonhint:
+    tooltip: "Called by lljson encode and slencode to determine if table should be encoded as an array or object."
```
tapple:
```suggestion
    tooltip: 'String property used by lljson.encode. Must be one of "array" or "object"."
```

slua_definitions.yaml
```diff
@@ -2812,3 +2812,48 @@ builtin-functions:
...
+  __call:
+    tooltip: "Method called when table is called as a function."
+  __concat:
+    tooltip: "Overload method for .. operator."
+  __div:
+    tooltip: "Overload method for / operator."
+  __eq:
+    tooltip: "Overload method for == operator."
+  __index:
+    tooltip: "Method called when table is indexed."
```
tapple:
```suggestion
    tooltip: "Method or table called when table is indexed."
```

slua_definitions.yaml
```diff
@@ -2814,3 +2812,48 @@ builtin-functions:
...
+  __metatable:
+    tooltip: "Override the table returned by getmetatable and cause setmetatable to raise an error."
+  __mod:
+    tooltip: "Overload method for % operator."
+  __mode:
+    tooltip: "Controls weak mapping of table valid returns are 'k' for keys and 'v' for values or 'kv' for both."
+  __mul:
+    tooltip: "Overload method for * operator."
+  __newindex:
+    tooltip: "Method called when table is indexed with an assignment."
```
Suzanna-Linn:
```suggestion
    tooltip: "Method or table called when table is indexed with an assignment."
```

lsl_definitions/generators/textmate.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def get_LL_module_functions(definitions: SLuaDefinitions) -> dict:
+        functions = [f"{f.name}" for f in definitions.get_module("ll").functions if not f.private and not f.local_only]
+        return crunch_regex_strings(functions)
+
+    def get_slua_global_types(definitions: SLuaDefinitions) -> str:
+        missing_types = ["nil","symbol", "userdata"]
+        builtin_types = [t for t in definitions.builtin_types.keys()]
+        base_classes = [f"{b.name}" for b in definitions.base_classes]
+        type_aliases = [f"{t.name}" for t in definitions.type_aliases]
```
tapple:
```suggestion
        type_aliases = [f"{t.name}" for t in definitions.type_aliases] if t.export
```

lsl_definitions/generators/textmate.py
```diff
@@ -0,0 +1,197 @@
...
+        modules = [m for m in modules if m is not None]
+        modules.sort()
+        return "|".join(modules)
+
+    def get_LL_module_functions(definitions: SLuaDefinitions) -> dict:
+        functions = [f"{f.name}" for f in definitions.get_module("ll").functions if not f.private and not f.local_only]
+        return crunch_regex_strings(functions)
+
+    def get_slua_global_types(definitions: SLuaDefinitions) -> str:
+        missing_types = ["nil","symbol", "userdata"]
```
tapple: - unsure what symbol is
- userdata is not a keyword, only a string that `type()` can return
- nil is duplicated in the builtinConstants section, but I don't think you use that section of SLuaDefinitions yet

lsl_definitions/generators/textmate.py
```diff
@@ -0,0 +1,197 @@
...
+    def get_slua_global_constants_for_module(module: SLuaModule) -> str:
+        constants = [f"{c.name}" for c in module.constants if not c.private]
+        constants.sort()
+        constants = "|".join(constants)
+        if len(constants) > 0:
+            constants = f"(\\.{constants})?"
+        return f"{module.name}{constants}"
+
+    def get_slua_global_constants(definitions: SLuaDefinitions) -> str:
+        constants = [get_slua_global_constants_for_module(m) for m in definitions.modules]
```
tapple, line 114: also add definitions.builtinConstants (true, false, nil)

slua_definitions.yaml
tapple: can `__newindex` be a table? unsure how that would work

lsl_definitions/generators/textmate.py
tapple: wait (true, false, nil) are already in the template, nevermind

lsl_definitions/generators/textmate.py
tapple: nil is also already in the template

slua_definitions.yaml
Suzanna-Linn:
```
local a = { "one", "two", "three" }
local b = {}
setmetatable(a, { __newindex = b })
for i, v in { "uno", "dos", "tres", "cuatro" } do
    a[i] = v
end
print(table.concat(a, ", "))
-- > uno, dos, tres
for k, v in b do print(k, v) end
-- > 4    cuatro
```

templates/syntax/slua.tmLanguage.json
```diff
@@ -0,0 +1,934 @@
...
+          {
+            "match": "\\b([a-zA-Z_][a-zA-Z0-9_]*)\\b",
+            "name": "variable.other.readwrite.luau"
+          }
+        ]
+      },
+      "operator": {
+        "patterns": [
+          {
+            "match": "==|~=|!=|<=?|>=?",
```
tapple:
```suggestion
            "match": "==|~=|<=?|>=?",
```

templates/syntax/slua.tmLanguage
```diff
@@ -0,0 +1,1434 @@
...
+          </dict>
+        </array>
+      </dict>
+      <key>operator</key>
+      <dict>
+        <key>patterns</key>
+        <array>
+          <dict>
+            <key>match</key>
+            <string>==|~=|!=|&lt;=?|&gt;=?</string>
```
tapple:
```suggestion
            <string>==|~=|&lt;=?|&gt;=?</string>
```

templates/syntax/slua.tmLanguage.json
tapple: - upstreamed also: https://github.com/JohnnyMorganz/Luau.tmLanguage/pull/27

lsl_definitions/generators/textmate.py
tapple: - upstreamed: https://github.com/JohnnyMorganz/Luau.tmLanguage/pull/28

lsl_definitions/generators/textmate.py
```diff
@@ -0,0 +1,236 @@
...
+        SLUA_GLOBAL_LL_VARIABLE_FUNCTIONS_REGEX=get_LL_variable_functions(slua_definitions),
+        SLUA_GLOBAL_LL_VARIABLES_REGEX=get_LL_variables(slua_definitions),
+    )
+
+    _base, ext = os.path.splitext(template_path)
+
+    with open(template_path, "r") as f:
+        template = Template(f.read())
+        if ext == ".tmLanguage":
+            return template.safe_substitute(inserts)
```
tapple, line 219: I'd prefer you used `substitute` over `safe_substitute`. It's more strict, and everything in this repo is meant to validate as strictly as possible and fail fast

lsl_definitions/generators/textmate.py
```diff
@@ -0,0 +1,236 @@
...
+
+@register("syntax_textmate_lsl")
+def gen_textmate_lsl(definitions: LSLDefinitions, template_path: str) -> str:
+
+    inserts = dict(
+        LSL_FLOW_CONTROL_REGEX=crunch_regex_strings(["default", "event","print"] + [c for c in definitions.controls.keys()]),
+        LSL_TYPES_REGEX=crunch_regex_strings([t for t in definitions.types.keys()]),
+        LSL_EVENTS_REGEX=crunch_regex_strings([e for e in definitions.events.keys()]),
+
+        LSL_FUNCTIONS_REGEX=crunch_regex_strings([f"{name}" for name, func in definitions.functions.items() if not func.private and not func.deprecated and not func.god_mode]),
```
tapple:
```suggestion
        LSL_FUNCTIONS_REGEX=crunch_regex_strings([name for name, func in definitions.functions.items() if not func.private and not func.deprecated and not func.god_mode]),
```

lsl_definitions/generators/textmate.py
```diff
@@ -0,0 +1,236 @@
...
+            inserts = {k: encode_json_unquoted(v) for k, v in inserts.items()}
+            return template.safe_substitute(inserts)
+        else:
+            raise ValueError(f"Unsupported template extension: {ext}")
+
+@register("syntax_textmate_lsl")
+def gen_textmate_lsl(definitions: LSLDefinitions, template_path: str) -> str:
+
+    inserts = dict(
+        LSL_FLOW_CONTROL_REGEX=crunch_regex_strings(["default", "event","print"] + [c for c in definitions.controls.keys()]),
```
tapple: 1. add `default` to `lsl_definitions.yaml` in the controls section, rather than here
2. I don't think `event` and `print` are valid lsl keywords

templates/syntax/lsl.tmLanguage
```diff
@@ -0,0 +1,398 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tapple, line 1: It looks to me like the 2 xml templates could be eliminated, and the json file could be used to generate both formats. But, probably better to do that in a follow-up PR so that it's easy to see what changed, and audit any differences that may exist between the two files that we don't catch here

lsl_definitions/generators/textmate.py
```diff
@@ -0,0 +1,236 @@
...
+from string import Template
+
+import os
+import json
+import re
+
+
+class Trie():
+    """Regex::Trie in Python. Creates a Trie out of a list of words. The trie can be exported to a Regex pattern.
+    The corresponding Regex should match much faster than a simple Regex union."""
```
tapple: did you get this from somewhere? a source url would be nice if so

gen_all_definitions.sh
```diff
@@ -3,7 +3,7 @@
 set -ex
```
tapple, line 3: Note in the PR description that this
- Closes #9

generated/syntax/slua.tmLanguage
```diff
@@ -0,0 +1,1446 @@
...
+            <key>name</key>
+            <string>support.function.luau</string>
+            <key>match</key>
+            <string>(?&lt;![^.]\.|:)\b(LL(?:Events:(?:eventNames|handlers|o(?:ff|n(?:ce)?))|Timers:(?:every|o(?:ff|nce))))\b</string>
+          </dict>
+          <dict>
+            <key>name</key>
+            <string>support.constant.luau</string>
+            <key>match</key>
+            <string>(?&lt;![^.]\.|:)\bbit32|buffer|coroutine|debug|ll|llbase64|llcompat|lljson(\.array_mt|empty_array|empty_object|null|object_mt|remove)?|llprim(\.ParamsSetter)?|math(\.huge|pi)?|os|quaternion(\.identity)?|rotation(\.identity)?|string|table|utf8(\.charpattern)?|uuid|vector(\.one|zero)?\b</string>
```
tapple, line 648: `llprim.ParamSetter` is in this list of constants. honestly, I don't think you can highlight it any better than that, for the same reason the syntax highlighter can't detect that `v.x` is a vector property access. It doesn't have a global singleton like `LLEvents` and `LLTimers`

The only other thing the highlighter could possibly highlight is `llprim.ParamSetter.new`, and, I don't think that's worth it the cross-refferencing it would take to generate that rule

tapple: you'd have to
1. check all constants
2. see if `c.value` or `c.value.removesuffix("Meta")` matches a class name
3. Add all functions (not methods) in that class to the function list

Step 2 requires not using `removesuffix` because Python3.7 is too old for it
Step 3 requires rebasing atop #119

All that just for 1 constructor function so far. will we add more? probably, but not a lot most likely. and the methods (the actual fun part) aren't highlightable ever

templates/syntax/lsl.tmLanguage
tapple: actually, there's one important difference between xml and json as template files: xml supports comments, and json doesn't. You commented out some parts of the xml template, but can't in the json template. So, if I had to pick one template over the other to keep, it would be the xml one because comments

templates/syntax/slua.tmLanguage
```diff
@@ -0,0 +1,1446 @@
...
+      </dict>
+      <dict>
+        <key>include</key>
+        <string>#string</string>
+      </dict>
+      <!-- leave shebang out for slua at the moment -->
+      <!--dict>
+        <key>include</key>
+        <string>#shebang</string>
+      </dict-->
```
tapple, line 61: Remove all the commented out sections from json template

slua_definitions.yaml
tapple: huh. I guess that makes sense. I may be able to use that to implement cascading dictionaries. Those were really annoying to implement in LSL

lsl_definitions/generators/textmate.py
Suzanna-Linn: `print` exists in LSL, but does nothing in-world:
```
default {state_entry() {
    print("Hello console");  // nothing visible
}}
```
It prints to the console, it's only usable by Lindens when debugging out-world.

WolfGangS: both are reserved keywords, they should probably go into some `illegal` or `deprecated` option instead of here though.

tapple: lol. I found an alchemy viewer crash with print

tapple: - #123
- #124

templates/syntax/lsl.tmLanguage
tapple: actually actually, since harold pointed out that python includes a plist generator in stdlib, I'd rather reformat this in yaml in a future PR. That's easier to work with than both xml and json, and matches luau upstream better (for LSL, I guess we are the upstream now)

lsl_definitions/generators/textmate.py
tapple: oh. nil needs to be in the textmate file in two different contexts: 
1. as a constant and 
3. as a type literal. 

it's fine to leave it in `missing_types` then. Either that, or put it in the template directly, same as `true` and `false` are. But `missing_types` makes for a smaller diff with upstream, which is preferable, imo

tapple: If you do keep `missing_types` (I would, due to smaller upstream diff), note the reasoning in a python comment

lsl_definitions/generators/textmate.py
WolfGangS: This would require converting all regex `$`'s  to `$$` in the templates, creating a bigger diff to upstream luau template. I don't object just want your thoughts.

Given that the plan is to move away from a template entirely it may not be so much of an issue.

lsl_definitions/generators/textmate.py
WolfGangS: Done

lsl_definitions/generators/textmate.py
WolfGangS: https://gist.github.com/EricDuminil/8faabc2f3de82b24e5a371b6dc0fd1e0

lsl_definitions/generators/textmate.py
WolfGangS: I'll leave them in for now, till the status of those pr's is resolved

templates/syntax/lsl.tmLanguage
WolfGangS: Agreed, once this pr is done, a phase 2 rework to generate from a yaml source would be good. and this work can give a baseline to compare against.

lsl_definitions/generators/textmate.py
tapple: oh that's fine then.

I have no plan to move away from templates.

lsl_definitions/generators/textmate.py
tapple: in the code please

lsl_definitions/generators/textmate.py
WolfGangS: Done

generated/syntax/slua.tmLanguage
WolfGangS: Ok, leaving this be for now then.

generated/syntax/lsl.tmLanguage
```diff
@@ -0,0 +1,398 @@
...
+		</dict>
+		<dict>
+			<key>include</key>
+			<string>#comments</string>
+		</dict>
+		<dict>
+			<key>match</key>
+			<string>==|!=</string>
+			<key>name</key>
+			<string>keyword.operator.comparison.lsl</string>
```
tapple, line 39: I'm curious at a high-level about how textmate works. It looks like these `name`s are probably css tags it adds to code that `match`es  the pattern. Are those tags well-known, and have well-known colors/styles? or is the color/style info somewhere else?

tapple (approved):

generated/syntax/lsl.tmLanguage
```diff
@@ -0,0 +1,398 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
```
tapple, line 2: Have you sanity checked these files in a textmate highlighter? I don't know how to do that

tapple (approved):

lsl_definitions/slua.py
```diff
@@ -327,13 +328,26 @@ class SLuaDefinitions:
...
     _TYPE_SEPERATORS_RE = re.compile(
         r"[ \n?&|,{}\[\]()]|\.\.\.|typeof|->|[a-zA-Z0-9_]*:|\"[a-zA-Z0-9_]*\""
     )
 
+    def get_module(self, name: str) -> SLuaModule:
+        for m in self.modules:
+            if m.name == name:
+                return m
+        return None
+
```
tapple: it won't pass CI/CD. Run this:
```
pre-commit run --all
```

lsl_definitions/generators/textmate.py
```diff
@@ -0,0 +1,269 @@
...
+
+        alt = []
+        cc = []
+        q = 0
+        for char in sorted(data.keys()):
+            if isinstance(data[char], dict):
+                try:
+                    recurse = self._pattern(data[char])
+                    alt.append(self.quote(char) + recurse)
+                except:
```
tapple: from pre-commit
```suggestion
                except KeyError:
```

lsl_definitions/slua.py
tapple: for future work, maybe do this so it fixes on commit:
```
pre-commit install
```

I modified `.git/hooks/pre-commit` so it skips the slow file generation check every commit attempt:
```
    SKIP=check-generated exec "$INSTALL_PYTHON" -mpre_commit "${ARGS[@]}"
```

generated/syntax/lsl.tmLanguage
tapple: It looks like these tags are indeed well-known: https://macromates.com/manual/en/language_grammars#naming_conventions

templates/syntax/lsl.tmLanguage.json
```diff
@@ -0,0 +1,277 @@
+{
+    "information_for_contributors": [
+      "The source of this file is https://github.com/JohnnyMorganz/Luau.tmLanguage/blob/main/Luau.YAML-tmLanguage",
+      "If you want to provide a fix or improvement, please create a pull request against the original repository.",
+      "Once accepted there, we are happy to receive an update request."
```
tapple: LSL contribution/attribution is incorrect

tapple: This is still incorrect. This file has nothing to do with JohnnyMorganz

tapple: This is still incorrect. JohnnyMorganz had nothing to do with LSL

generated/syntax/lsl.tmLanguage
```diff
@@ -2,6 +2,19 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+    <key>information_for_contributors</key>
+    <array>
+      <string>The source of this file is https://github.com/secondlife/lsl-definitions/tree/main/generated/syntax/lsl.tmLanguage</string>
+      <string>If you want to provide a fix or improvement, please create a pull request against the original repository.</string>
+      <string>Once accepted there, we are happy to receive an update request.</string>
```
tapple: This is the original repository now

templates/syntax/lsl.tmLanguage.json
WolfGangS: Sorry, i spotted a problem with the contribution info in the slua files at the same time and skimmed your last comment believing it to be the same issue.

Should both be fixed now

generated/syntax/lsl.tmLanguage
WolfGangS: The Slua one is already in use here, https://slua-playground.wlf.io

Tested the lsl one mostly using https://www.leskoff.com/s02050-0  probably need a cleaner solution but that was fast an simple :stuck_out_tongue_closed_eyes:

templates/syntax/LICENSE.md
```diff
@@ -0,0 +1,47 @@
+# Attributions
+
+## lsl.tmLanguage
+
+Was adapted from
+
+ - https://github.com/textmate/secondlife-lsl.tmbundle
+
```
tapple: pre-commit complaining about whitespace here
```suggestion

```

lsl_definitions/generators/textmate.py
```diff
@@ -0,0 +1,314 @@
...
+## Trie Code
+# Original Perl module: Regexp::Trie
+# Original Copyright (C) 2006 by Dan Kogai
+#
+# This Python translation is a derivative work based on Regexp::Trie
+# Copyright (c) 2010 by rex
+# Copyright (c) 2017 by fcicq, atiking and EricDuminil
+# This Python code is licensed under the Artistic License 2.0 (AL2.0).
+# A copy of the Artistic License 2.0 can be found at:
+# https://opensource.org/licenses/Artistic-2.0
```
HaroldCindy: This may be a sticking point, unfortunately. The Artistic License 2.0, while a reasonable license, is more strict than the MIT license that covers this repo. The regex trie implementation looks pretty trivial, essentially uses nested non-capturing groups to branch on shared prefixes from a sorted string list? Might be preferable to just rewrite it.

lsl_definitions/generators/textmate.py
```diff
@@ -0,0 +1,314 @@
...
+    rotation_module = SLuaModule(
+        name="rotation",
+        comment=quaternion_module.comment,
+        callable=quaternion_module.callable,
+        constants=quaternion_module.constants,
+        functions=quaternion_module.functions,
+    )
+    slua_definitions.modules.append(rotation_module)
+
+    def get_slua_global_functions(definitions: SLuaDefinitions) -> dict:
```
HaroldCindy: Says it returns a dict, returns a string.

We should probably get some `mypy` in the repo, now that I think about it.

lsl_definitions/generators/textmate.py
```diff
@@ -0,0 +1,314 @@
...
+
+    _base, ext = os.path.splitext(template_path)
+
+    with open(template_path, "r") as f:
+        template = Template(f.read())
+        if ext == ".tmLanguage":
+            return template.safe_substitute(inserts)
+        elif ext == ".json":
+            inserts = {k: encode_json_unquoted(v) for k, v in inserts.items()}
+            return template.safe_substitute(inserts)
```
HaroldCindy, line 273: Using string templating to pull together JSON is a little wonky, would it be more sensible to parse the `.json` and shove things into well-known paths within the JSON?

lsl_definitions/slua.py
```diff
@@ -334,6 +335,18 @@ class SLuaDefinitions:
         r"[ \n?&|,{}\[\]()]|\.\.\.|typeof|->|[a-zA-Z0-9_]*:|\"[a-zA-Z0-9_]*\""
     )
 
+    def get_module(self, name: str) -> SLuaModule:
```
HaroldCindy, line 338: Hmm. Makes me wonder if `List` is even the right shape for these members, but not a blocker. @tapple any thoughts?

templates/syntax/LICENSE.md
```diff
@@ -0,0 +1,47 @@
...
+## slua.tmLanguage and slua.tmLangauge.json
+
+Most of these files was addapted from the works below.
+ - https://github.com/JohnnyMorganz/Luau.tmLanguage
+ - https://github.com/NightrainsRbx/RobloxLsp/blob/master/syntaxes/lua.tmLanguage.json
+
+
+### Licenses
+
+Copyright (c) JohnnyMorganz All rights reserved.
```
HaroldCindy, line 28: The easiest way to comply with the license is to shove them into comments in the output `.tmLanguage`, rather than just linking them. I believe it would be sufficient to say "Based on <x>, Copyright (year) (so and so) - MIT license" in the relevant plist comments since the body of the MIT license is already shipped with this repo.

slua_definitions.schema.json
```diff
@@ -327,6 +327,24 @@
                 }
             },
             "type": "object"
+        },
+        "metamethods": {
```
HaroldCindy, line 331: nice, thanks for this!

slua_definitions.yaml
```diff
@@ -2836,3 +2836,48 @@ builtin-functions:
...
+    tooltip: "Method called by tostring to convert the value to a string."
+  __unm:
+    tooltip: "Overload method for - as unaryoperator."
+  __iter:
+    tooltip: "Method called to iterate over the table."
+  __idiv:
+    tooltip: "Overload method for // operator."
+  __tojson:
+    tooltip: "Called by lljson encode and slencode to encode the value as JSON."
+  __jsonhint:
```
HaroldCindy, line 2882: This isn't strictly a metamethod. Functionally speaking it's a random property that we just give an `__` prefix to clarify that it's a SLua-internal concern. Maybe doesn't need to be specifically highlighted, I'm not sure.

lsl_definitions/generators/textmate.py
```diff
@@ -0,0 +1,314 @@
...
+        else:
+            raise ValueError(f"Unsupported template extension: {ext}")
+
+
+@register("syntax_textmate_lsl")
+def gen_textmate_lsl(definitions: LSLDefinitions, template_path: str) -> str:
+    inserts = dict(
+        LSL_FLOW_CONTROL_REGEX=crunch_regex_strings([c for c in definitions.controls.keys()]),
```
HaroldCindy: can simply be `list(definitions.yadayada.keys())` if you specifically need a list, the comprehension makes it look like there's some other logic going on there, and there isn't.

HaroldCindy (changes requested):
Thanks for taking this on!

In general, it looks alright. There's certainly a lot of wonkiness with the nested functions inside other functions, but that's not anything we don't do elsewhere.

I imagine some of that could be abstracted out but... this is a repo for (mostly) straightforward codegen with deterministic outputs. So long as the generated code looks fine, the codegen code can have some warts :)

I do have a few comments though, lmk if there's anything I can clarify!

lsl_definitions/generators/textmate.py
tapple: rewrote referencing only wikipedia: https://github.com/secondlife/lsl-definitions/pull/120/changes/7523f39365cfc51a49016a34a99fbf3b5508e81c

lsl_definitions/generators/textmate.py
tapple: maybe I'll do that during a rewrite #132. At the moment, the json/xml template is not parsed

lsl_definitions/slua.py
tapple: No strong opinion. It was a list before I got here, and I left it that way. It helped some when I was creating the initial parsers and needed to compare against an upstream file that wasn't sorted. That's why some slua functions are not in alphabetical order

tapple: Some other generators also do this. WolfGang just shortened it

https://github.com/secondlife/lsl-definitions/blob/ce168d38539668f78c0bc8bb14b575c7daa1fa12/lsl_definitions/generators/slua.py#L25-L26

https://github.com/secondlife/lsl-definitions/blob/ce168d38539668f78c0bc8bb14b575c7daa1fa12/lsl_definitions/generators/slua.py#L243

slua_definitions.yaml
tapple: It's not the only one that isn't a function. I'm fine with calling them metamethods anyway:
- `__metatable` is a table
- `__mode` is a string, just like `__jsonhint`
- `__index` and `__newindex` can be either functions or tables

templates/syntax/LICENSE.md
tapple: It was like that. WolfGang changed it as the attribution was getting a bit long: https://github.com/secondlife/lsl-definitions/pull/120/changes/c645623c802995847ef1cdfe2993239aee334353. I'm happy either way

lsl_definitions/generators/textmate.py
WolfGangS: It could be done i went for simple text insert template for json, to match the xml for now.

Paths are a slight problem as some of the items are in lists and I'm not sure about hard coding to indexes that have no "inherent" meaning. Would need to loop lists and insert on match.

I agree its probably the right way, but maybe for a rewrite, lets get a working implementation out first, then have something to judge a more complex rewrite against.

lsl_definitions/generators/textmate.py
WolfGangS: Marking as resolved, Thank you @tapple

slua_definitions.yaml
HaroldCindy: Fair enough!

lsl_definitions/generators/textmate.py
HaroldCindy: Yeah I suppose so, since this is just static codegen and you don't have to care too much about user-provided data creeping in and corrupting things.

generated/syntax/lsl.tmLanguage
```diff
@@ -105,15 +105,15 @@
 			<key>comment</key>
 			<string>LSL Events</string>
 			<key>match</key>
-			<string>\b(?:at(?:_(?:rot_target|target)|tach)|c(?:hanged|o(?:llision(?:_(?:end|start))?|ntrol))|dataserver|e(?:mail|xperience_permissions(?:_denied)?)|final_damage|game_control|http_re(?:quest|sponse)|l(?:and_collision(?:_(?:end|start))?|i(?:nk(?:_message|set_data)|sten))|mo(?:ney|ving_(?:end|start))|no(?:_sensor|t_at_(?:rot_target|target))|o(?:bject_rez|n_(?:d(?:amage|eath)|rez))|path_update|r(?:emote_data|un_time_permissions)|s(?:ensor|tate_e(?:ntry|xit))|t(?:imer|ouch(?:_(?:end|start))?|ransaction_result))\b</string>
+			<string>\b(?:at(?:_(?:rot_target|target)|tach)|c(?:hanged|o(?:llision(?:|_(?:end|start))|ntrol))|dataserver|e(?:mail|xperience_permissions(?:|_denied))|final_damage|game_control|http_re(?:quest|sponse)|l(?:and_collision(?:|_(?:end|start))|i(?:nk(?:_message|set_data)|sten))|mo(?:ney|ving_(?:end|start))|no(?:_sensor|t_at_(?:rot_target|target))|o(?:bject_rez|n_(?:d(?:amage|eath)|rez))|path_update|r(?:emote_data|un_time_permissions)|s(?:ensor|tate_e(?:ntry|xit))|t(?:imer|ouch(?:|_(?:end|start))|ransaction_result))\b</string>
```
HaroldCindy, line 108: Ahhhh I see, the new regex trie implementation handles a conditional suffix through `(?:|something)` rather than `(?:something)?`. Struck me as odd at first glance, but I think it's fine so long as either works, thanks for this!

templates/syntax/LICENSE.md
HaroldCindy: Yep, the MIT license is fairly explicit about this. If we ever give someone a copy of the `tmLanguage` file separately from this repo (which we will, through the JSON-RPC bridge), we need to include a copy of the full MIT license and the relevant copyright holders to be compliant.

It's why you get dozens of pages about zlib and curl license information in the credits of various games.

HaroldCindy (approved):
LGTM, thanks! Just one pending note about including the third-party license details in the files where they're relevant and this should be good to merge.

templates/syntax/LICENSE.md
tapple: added: https://github.com/secondlife/lsl-definitions/pull/120/changes/d2df52c52c93f4fc612ea960a7a7d68492865625

It only seemed to be a requirement of the slua files, not lsl files, so I didn't change the lsl files

generated/syntax/lsl.tmLanguage
tapple: I went for compactness of the trie code over compactness of the regex

PR #123, opened 2026-05-09
Move LSL keyword `default` to `lsl_definitions.yaml`

tapple:
`default` was misplaced in the keywords file as a constant, rather than a control, and the generators had some special handling for it. Move it to the logical place, and remove the exceptions

This also removes exceptional handling for it in the unmerged textmate generator: https://github.com/secondlife/lsl-definitions/pull/120#pullrequestreview-4255198723

This changes it's in-viewer highlighting color from blue to purple


HaroldCindy:
It's really only under there so the color of `default` is distinct from `state foo`. Is having it in `constants` breaking anything?

tapple:
it's not breaking anything, in that:
1. it's been this way since time immemorial
2.  it's only one item in the wrong place and thus is easy to single out an exception for.

Why somebody wanted it to have a different color than `state foo`, I have no idea (I would prefer they matched), but, that's the only theory present in the code.

The motivation for this PR is indeed weak: removing a one-line exception from the textmate generator, and 7 lines of exception elsewhere

While I don't have a strong opinion about the highlighting color (bikeshed topic), it would be nice to move `default`'s description source from `lsl.py` to `lsl_definitions.yaml` (this would also get rid of the textmate exception). I could still put it in with the constants in the `keywords.xml` generator, to preserve the color

WolfGangS:
Personally i agree it should be in flow control.

But i do think it needs a special case to be kept in constants for now for `keywords.xml` changing that after so many years is probably not advisable.

I'm pretty certain that loads of less familiar scripters would think its an error or just miss it being blue :cry:

tapple:
moved it back to `constants` in `keywords.xml`: https://github.com/secondlife/lsl-definitions/pull/123/changes/a9596142261575f85fcf60d62d1d7c84a3239dd5. Now this PR no longer changes `keywords.xml`

HaroldCindy (approved):

PR #124, opened 2026-05-09
Document the LSL reserved words `print` and `event`

tapple:
Add the lsl reserved words `print` and `event` to the controls section, marked as `deprecated`.

I did not mark them as `private` because they currently cause unexplained syntax errors if you use them as variable names. Therefore, it's preferable hilight them a funny color and give them a tooltip with some context. 

`event` now renders like this in the viewer. Marking it `deprecated` colors it pink, even though this was a previously unused code path for `controls` keywords in the viewer.:
lsl-event

See also:
- https://github.com/secondlife/lsl-definitions/pull/120#pullrequestreview-4255198723


lsl_definitions.yaml
```diff
@@ -15807,13 +15807,19 @@ controls:
...
+    deprecated: true
+    tooltip: Undefined reserved word.
   for:
     tooltip: for loop\nfor (<initializer>; <condition>; <post-iteration-statement>)\n{
       ...\n}
   if:
     tooltip: if / else block\nif (<condition>) {\n...\n[} else [if (<condition>) {\n...]]\n}
   jump:
     tooltip: jump statement\njump <label>
+  print:
```
tapple, line 15820: I tried to put `print` in the functions section, since it is, but, encountered several problems doing so:
1. Numerous places in the code require all functions to start with `ll`, and this one doesn't.
2. This function (probably) doesn't have a `func-id`, and the validator requires one.
3. It's argument has a totally unique typing: it accepts any argument type without error (string, integer, float, list, vector, quaternion, key). No other LSL function behaves that way, and the schema has no way to indicate this.
4. This function doesn't exist in SLua

Therefore, I assumed this is probably a compiler builtin of some sort, gave up, and moved it to the `controls` section. I did not try to check any lsl compiler source code to validate this assumption.

tapple:
```luau
print(ll.print) -- nil
```
For a code example, see: https://feedback.secondlife.com/slua-alpha/p/undocumented-lsl-print-function-works-differently-under-lsl-2025-vm

tapple: Since it sounds like Harold intentionally improved `print` (https://feedback.secondlife.com/slua-alpha/p/undocumented-lsl-print-function-works-differently-under-lsl-2025-vm), I could move it from controls to functions as part of the `keywords.xml` generator (similarly to the alternate proposal in https://github.com/secondlife/lsl-definitions/pull/123#issuecomment-4414053348). Though, that might cause trouble for non-viewer consumers of `keywords.xml`

generated/lsl_keywords_pretty.xml
```diff
@@ -30,6 +37,13 @@ This file is auto-generated by https://github.com/secondlife/lsl-definitions. --
             <key>tooltip</key>
             <string>jump statement\njump &lt;label&gt;</string>
          </map>
+         <key>print</key>
+         <map>
+            <key>deprecated</key>
+            <boolean>true</boolean>
+            <key>tooltip</key>
+            <string>Prints text to stdout on the server. For internal use only.</string>
```
HaroldCindy: It's not really just reserved for internal use any more, but it's only meaningful under VM 2025 or Lua

generated/lsl_keywords_pretty.xml
```diff
@@ -30,6 +37,13 @@ This file is auto-generated by https://github.com/secondlife/lsl-definitions. --
             <key>tooltip</key>
             <string>jump statement\njump &lt;label&gt;</string>
          </map>
+         <key>print</key>
+         <map>
+            <key>deprecated</key>
+            <boolean>true</boolean>
```
HaroldCindy: Not deprecated

tapple: fixed

generated/lsl_keywords_pretty.xml
tapple: Updated tooltip

HaroldCindy (approved):

PR #126, opened 2026-05-12
add dataserver-semantics

tapple:


lsl_definitions.yaml
```diff
@@ -15494,6 +15522,9 @@ functions:
     experience: true
     func-id: 389
     return: key
+    dataserver-semantics: csv
+    - boolean
+    -fridge
```
Suzanna-Linn, line 15527: type `fridge`...

Can a key have value and list? I would say that no, but I'm not sure.

tapple: no, I don't think this is valid yaml. it's just notes, not a proposal yet. I wanted to improve how types and type semantics are handled before I finish this, but, that's waiting on approval of #113

tapple:
Closing in favor of #185

Issue #122, opened 2026-05-08
Create a generator for `EmbeddedBuiltinDefinitions.cpp`


tapple:
This C file is where slua gets its globals. `slua-analyze` won't work without it unless it grows support for `d.luau` files sometime soon

There's 2 places that would be really nice to typecheck, where including luau-lsp would be vastly overkill:
- slua playground https://slua-playground.wlf.io/
- the secondlife viewer

But, this is impractical without a generator for  `EmbeddedBuiltinDefinitions.cpp`

---

One other benefit of embedding the slua definitions has to do with the magic type functions. The short version: due to the order in which `luau-lsp` reads the definitions, `luau-lsp` is unable to use the best available typechecker logic for the functions `table.freeze`, `table.clone`, and `table.pack`. For the long version, see these comments, and the whole PR they are in:
- [secondlife.d.luau shadows the `table` magic functions](https://github.com/secondlife/lsl-definitions/pull/130#pullrequestreview-4369502372)
- [`table.freeze` magic function](https://github.com/secondlife/lsl-definitions/pull/130#discussion_r3301494868)
- [`table.clone` magic function](https://github.com/secondlife/lsl-definitions/pull/130#discussion_r3248443423)
- [`table.pack` magic function](https://github.com/secondlife/lsl-definitions/pull/130#discussion_r3249128744)

CatfishW:
/try

Plan:
- add a real `EmbeddedBuiltinDefinitions.cpp` generator path instead of another partial helper
- generate the full builtin source from both the Luau builtin section and the SLua/library section in `slua_definitions.yaml`
- emit a drop-in C++ file via a template, wire it into `gen_all_definitions.sh`, and add the generated output under `generated/cpp/` for verification
- keep the existing `slua_lsp_defs` output stable while adding the new generator

Rider-Linden:
On the viewer side, it receives an updated d.luau (along with any other config files) from the simulator as soon as the user opens a script. The lua viewer will also have it's own copy of this file when packaged. 

I can't speak to slua-playground.

PR #130, opened 2026-05-13
Mark the functions that have a magic type

tapple:
- Follows #133 
- Part of #122 

Mark 2 classes of functions that need special handling in the generators of `secondlife.d.luau` and `EmbeddedBuiltinDefinitions.cpp`
- builtin: functions not defined in [`EmbeddedBuiltinDefinitions.cpp`](https://github.com/secondlife/slua/blob/main/Analysis/src/EmbeddedBuiltinDefinitions.cpp) must be left out of both files
- magic type functions: Custom logic defined in [`BuiltinDefinitions.cpp`](https://github.com/secondlife/slua/blob/main/Analysis/src/BuiltinDefinitions.cpp) must be left out of only `secondlife.d.luau`. These all use `attachMagicFunction()`, hence the name "magic"

I also wanted to document what all the magic type functions do, and this PR seems like the most appropriate place to host that documentation for now


generated/secondlife.d.luau
```diff
@@ -451,14 +451,14 @@ declare quaternion: ((x: number, y: number, z: number, s: number) -> quaternion)
 declare string: {
   byte: (s: string, i: number?, j: number?) -> ...number,
   char: (...number) -> string,
-  find: (s: string, pattern: string, init: number?, plain: boolean?) -> (number?, number?, ...string),
-  format: (formatstring: string, ...any) -> string,
-  gmatch: (s: string, pattern: string) -> () -> ...string,
+  find: (s: string, pattern: string, init: number?, plain: boolean?) -> (number?, number?, ...string), -- magic type
```
tapple:
```luau
--!strict
local _start, _finish, _tag, _text, _extra = string.find(
    'monospace:: <code>print("hello regex")</code>. hooray',
    "<(%w-)>(.-)</%1>")
```

Magic detects the extra `string.find` variable assignment:
```
$ luau-lsp analyze --platform standard magic_string_find.luau 
[WARN] No definitions file provided by client
magic_string_find.luau(2,1): TypeError: Function only returns 4 values, but 5 are required here
```
Without magic, the extra parameter passes silently:
```
$ luau-lsp analyze --platform standard --defs secondlife.d.luau magic_string_find.luau 
[INFO] Loading definitions file: @roblox - sl-vscode-plugin/slua_fixes.d.luau
```

generated/secondlife.d.luau
```diff
@@ -451,14 +451,14 @@ declare quaternion: ((x: number, y: number, z: number, s: number) -> quaternion)
...
+  format: (formatstring: string, ...any) -> string, -- magic type
```
tapple:
```luau
--!strict
print(string.format("I am %s. I am %d years old.", "Pippin", "six"))
```
Magic knows that the 3rd `string.format` parameter should be a number:
```
$ luau-lsp analyze --platform standard magic_string_format.luau 
[WARN] No definitions file provided by client
magic_string_format.luau(2,62): TypeError: Expected this to be 'number', but got 'string'
```
Without magic, it passes silently:
```
$ luau-lsp analyze --platform standard --defs secondlife.d.luau magic_string_format.luau 
[INFO] Loading definitions file: @roblox - secondlife.d.luau
```

generated/secondlife.d.luau
```diff
@@ -451,14 +451,14 @@ declare quaternion: ((x: number, y: number, z: number, s: number) -> quaternion)
...
+  gmatch: (s: string, pattern: string) -> () -> ...string, -- magic type
```
tapple:
```luau
--!strict
for tag, text, extra in string.gmatch(
    'monospace:: <code>print("hello regex")</code>. hooray',
    "<(%w-)>(.-)</%1>"
) do
    print(tag, text, extra)
end
```
Magic detects the extra `string.gmatch` variable assignment:
```
$ luau-lsp analyze --platform standard magic_string_gmatch.luau 
[WARN] No definitions file provided by client
magic_string_gmatch.luau(2,1): TypeError: Argument count mismatch. Function expects 3 arguments, but only 2 are specified
```
Without magic, the extra parameter passes silently:
```
$ luau-lsp analyze --platform standard --defs secondlife.d.luau magic_string_gmatch.luau 
[INFO] Loading definitions file: @roblox - secondlife.d.luau
```

generated/secondlife.d.luau
```diff
@@ -451,14 +451,14 @@ declare quaternion: ((x: number, y: number, z: number, s: number) -> quaternion)
...
   gsub: (s: string, pattern: string, repl: string | { [string]: string } | (...string) -> string, maxn: number?) -> (string, number),
   len: (s: string) -> number,
   lower: (s: string) -> string,
-  match: (s: string, pattern: string, init: number?) -> ...string,
-  pack: (fmt: string, ...any) -> string,
+  match: (s: string, pattern: string, init: number?) -> ...string, -- magic type
```
tapple:
```luau
--!strict
local _tag, _text, _extra = string.match(
    'monospace:: <code>print("hello regex")</code>. hooray',
    "<(%w-)>(.-)</%1>")
```
Magic detects the extra `string.match` variable assignment:
```
$ luau-lsp analyze --platform standard magic_string_match.luau 
[WARN] No definitions file provided by client
magic_string_match.luau(2,1): TypeError: Function only returns 2 values, but 3 are required here
```
Without magic, the extra parameter passes silently:
```
$ luau-lsp analyze --platform standard --defs secondlife.d.luau magic_string_match.luau 
[INFO] Loading definitions file: @roblox - secondlife.d.luau
```

generated/secondlife.d.luau
```diff
@@ -515,9 +515,9 @@ declare table: {
   find: <V>(t: {V}, v: V, i: number?) -> number?,
   clear: (t: {}) -> (),
   shrink: <V>(t: {V}, shrink_sparse: boolean?) -> {V},
-  freeze: <table>(t: table) -> table,
+  freeze: <table>(t: table) -> table, -- magic type
   isfrozen: (t: {}) -> boolean,
-  clone: <table>(t: table) -> table,
+  clone: <table>(t: table) -> table, -- magic type
```
tapple, line 522:
```luau
--!strict
type objTable = {x: string}
type hashTable = {[string]: string}

local o: objTable = {x="yo"}
local _o: objTable = table.clone(o)

local h: hashTable = {x="yo"}
local _h: hashTable = table.clone(h)

local _n: number = table.clone(5)
```
Magic knows that `table.clone` only works on tables:
```
$ luau-lsp analyze --platform standard magic_table_clone.luau 
[WARN] No definitions file provided by client
magic_table_clone.luau(11,32): TypeError: Expected this to be '{-  -}', but got 'number'
magic_table_clone.luau(11,1): TypeError: Expected this to be 'number', but got '{-  -}'
```
Without magic, "generic table" is an unexpressable type and the file must resort to "generic anything"
```
$ luau-lsp analyze --platform standard --defs secondlife.d.luau magic_table_clone.luau 
[INFO] Loading definitions file: @roblox - secondlife.d.luau
```
`clone` magic only works on old solver:
```
$  luau-lsp analyze --platform standard --flag LuauSolverV2=true magic_table_clone.luau 
[WARN] No definitions file provided by client
```

generated/secondlife.d.luau
```diff
@@ -515,9 +515,9 @@ declare table: {
   find: <V>(t: {V}, v: V, i: number?) -> number?,
   clear: (t: {}) -> (),
   shrink: <V>(t: {V}, shrink_sparse: boolean?) -> {V},
-  freeze: <table>(t: table) -> table,
+  freeze: <table>(t: table) -> table, -- magic type
```
tapple, line 520:
```luau
--!strict
type objTable = {x: string}
type hashTable = {[string]: string}

local o: objTable = {x="yo"}
local _o: objTable = table.freeze(o)
_o.x = "error"

local h: hashTable = {x="yo"}
local _h: hashTable = table.freeze(h)
_h.x = "error"

local _n: number = table.freeze(5)
```
Magic knows that `table.freeze` only works on tables:
```
$ luau-lsp analyze --platform standard magic_table_freeze.luau 
[WARN] No definitions file provided by client
magic_table_freeze.luau(13,33): TypeError: Expected this to be '{-  -}', but got 'number'
magic_table_freeze.luau(13,1): TypeError: Expected this to be 'number', but got '{-  -}'
```
Without magic, "generic table" is an unexpressable type and the file must resort to "generic anything"
```
$ luau-lsp analyze --platform standard --defs secondlife.d.luau magic_table_freeze.luau 
[INFO] Loading definitions file: @roblox - secondlife.d.luau
```
Neither one understands the real purpose of `freeze` is to make a read-only table:
```
$ luau magic_table_freeze.luau 
./magic_table_freeze.luau:7: attempt to modify a readonly table
stacktrace:
./magic_table_freeze.luau:7
```

generated/secondlife.d.luau
```diff
@@ -508,7 +508,7 @@ declare table: {
   extend: <V>(a: {V}, b: {V}) -> {V},
   remove: <V>(a: {V}, i: number?) -> V?,
   sort: <V>(a: {V}, f: ((a: V, b: V) -> boolean)?) -> (),
-  pack: <V>(...V) -> { n: number, [number]: V },
+  pack: <V>(...V) -> { n: number, [number]: V }, -- magic type
```
tapple, line 513:
```luau
--!strict
local _: {string | number} = table.pack("a", 5)
```
Magic knows to union all the `table.pack` parameters for the return type:
```
$ luau-lsp analyze --platform standard magic_table_pack.luau 
[WARN] No definitions file provided by client
```
Without magic, this is not supported by the type system:
```
$ luau-lsp analyze --platform standard --defs secondlife.d.luau magic_table_pack.luau 
[INFO] Loading definitions file: @roblox - secondlife.d.luau
magic_table_pack.luau(2,46): TypeError: Expected this to be 'string', but got 'number'
```

generated/secondlife.d.luau
```diff
@@ -247,10 +247,10 @@ declare function rawequal(a: any, b: any): boolean
 declare function rawget<K, V>(t: {[K]: V}, k: K): V?
 declare function rawlen<K, V>(t: {[K]: V} | string): number
 declare function rawset<K, V>(t: {[K]: V}, k: K, v: V): {[K]: V}
-declare function require(target: string): any
-declare function select(i: string | number, ...any): any
+declare function require(target: string): any -- magic type
+declare function select(i: string | number, ...any): any -- magic type
 declare setfenv: nil
-declare function setmetatable(t: { [any]: any }, mt: { [any]: any }?): ()
+declare function setmetatable(t: { [any]: any }, mt: { [any]: any }?): () -- magic type
```
tapple: In the new solver, `setmetatable` does not need a magic type function to work, as it instead can be fully defined based on the builtin type function `setmetatable' (fixed in #135):
```
declare function setmetatable<T, MT>(t: T, mt: MT): setmetatable<T, MT>
```

In the old solver, `setmetatable` has both
1. an unexpressable builtin type
2. a magic type function

It's not possible to untangle those and demonstrate the builtin type without the magic type function (not without editing slua and recompiling), so I didn't make example code here to demonstrate it

HaroldCindy:
What's "magic"? I'm assuming it's an internal `luau-analyze` thing that does dynamic dispatch on how specifically to check the type based on the arguments passed. Is there a better term for it (i.e. whose magic is it anyway)?

slua_definitions.schema.json
```diff
@@ -187,6 +187,11 @@
                     "description": "See https://github.com/secondlife/slua/blob/main/Common/include/Luau/Bytecode.h#L542",
                     "type": "boolean"
                 },
+                "magic-type": {
+                    "const": true,
+                    "markdownDescription": "The typechecker has custom logic for this function. Look for attachMagicFunction in the source code.",
```
tapple: The meaning of "magic" is documented here. It's literally what the typechecker calls it

HaroldCindy: Yeah that makes sense, I just think it should just be obvious to people looking at the YAML what it's for without having to reference the schema (I definitely didn't notice it 😨). lsl-definitions has multiple output targets, so we could be a little clearer about what this affects. We needn't necessarily use the upstream terminology in our YAML... maybe something like `dynamic-typechecking`?

tapple: slua_definitions.yaml should probably have a big comment about what all the fields are like lsl_definitions.yaml has

I actually need another field, specifically for the `EmbeddedBuiltinDefinitions.cpp` generator, that means "this function has a type defined in c++, not in luau". It would apply to setmetatable and getmetatable, and the string table. The main reason for it is they have different type declarations under old and new solver. I was thinking of calling this field `c++_type` or `builtin-type`

tapple: I could combine these 2-3 boolean fields into one dict field, since they are all quite technical and related to typechecker internals:
```yaml
- typechecker:
  - magic: true
  - builtin: true
  - checked: true
```
`checked` would be from the next PR in this series: #125

generated/secondlife.d.luau
tapple: Luau 0.720 improved the `table.freeze` magic so that it can warn about read-only access:
```luau
--!strict
type objTable = {x: string}
local o: objTable = {x="yo"}
local of = table.freeze(o)
of.x = "error"
```
Magic now knows the table is read-only (new solver only):
```
$ luau-lsp analyze --platform standard --flag LuauSolverV2=true magic_table_freeze_bug.luau 
[WARN] No definitions file provided by client
magic_table_freeze_bug.luau(5,1): TypeError: Property x of table 'objTable' is read-only
```
Without magic, or in the old solver, table writing is not detected as an error:
```
$ luau-lsp analyze --platform standard --flag LuauSolverV2=true --defs secondlife.d.luau magic_table_freeze_bug.luau 
[INFO] Loading definitions file: @roblox - secondlife.d.luau
$ luau-lsp analyze --platform standard --flag LuauSolverV2=false magic_table_freeze_bug.luau 
[WARN] No definitions file provided by client
```

slua_definitions.schema.json
tapple: changed to dict: https://github.com/secondlife/lsl-definitions/pull/130/changes/f1e3c65f65fbd59d0ed9828a5aa7d1f18d302ea0

generated/secondlife.d.luau
```diff
@@ -230,27 +230,27 @@ export type PrimParamsSetterType = typeof(
 )
 
 
-declare function assert<T>(value: T?, message: string?): T
+-- declare function assert<T>(value: T?, message: string?): T -- magic type
```
tapple:
```luau
--!strict
local _a: string? = "hello"
assert(_a ~= nil)
local _b: string = _a
```
Magic lets `assert` do type narrowing:
```
$ luau-lsp analyze --platform standard magic_assert.luau 
[WARN] No definitions file provided by client
```
Without magic, no type narrowing happens:
```
$ luau-lsp analyze --platform standard --defs secondlife.d.luau magic_assert.luau 
[INFO] Loading definitions file: @roblox - secondlife.d.luau
magic_assert.luau(4,1): TypeError: Expected this to be 'string', but got 'string?'
```

generated/secondlife.d.luau
```diff
@@ -230,27 +230,27 @@ export type PrimParamsSetterType = typeof(
...
+-- declare function pairs<K, V>(t: {[K]: V}): (({[K], V}, K) -> (K?, V), {[K], V}, K) -- builtin
 declare function pcall<A..., R...>(f: (A...) -> R..., A...): (boolean, R...)
 declare function print(...any): ()
 declare function rawequal(a: any, b: any): boolean
 declare function rawget<K, V>(t: {[K]: V}, k: K): V?
 declare function rawlen<K, V>(t: {[K]: V} | string): number
 declare function rawset<K, V>(t: {[K]: V}, k: K, v: V): {[K]: V}
-declare function require(target: string): any
-declare function select(i: string | number, ...any): any
+-- declare function require(target: string): any -- magic type
```
tapple: ` magic_require.luau`
```luau
--!strict
local mod = require("magic_require_module")
local _nope = require("non_existent_module")
local _x: number = mod.x
local _hello: string = mod.world
```
` magic_require_module.luau`
```luau
--!strict
return { x = 1, y = 2 }
```
Magic lets `require` copy types between files:
```
$ luau-lsp analyze --platform standard magic_require.luau 
[WARN] No definitions file provided by client
magic_require.luau(3,15): TypeError: Unknown require: non_existent_module.lua
magic_require.luau(5,24): TypeError: Key 'world' not found in table '{ x: number, y: number }'
```
Without magic, `require` never errors:
```
$ luau-lsp analyze --platform standard --defs secondlife.d.luau magic_require.luau 
[INFO] Loading definitions file: @roblox - secondlife.d.luau
```

generated/secondlife.d.luau
```diff
@@ -230,27 +230,27 @@ export type PrimParamsSetterType = typeof(
...
+-- declare function select(i: string | number, ...any): any -- magic type
```
tapple:
```luau
--!strict
local _s: number = select(2, 5, "hello")
```
Magic knows the return type of `select` for each argument:
```
$ luau-lsp analyze --platform standard magic_select.luau 
[WARN] No definitions file provided by client
magic_select.luau(2,1): TypeError: Expected this to be 'number', but got 'string'
```
Without magic, this is not flagged by the type system:
```
$ luau-lsp analyze --platform standard --defs secondlife.d.luau magic_select.luau 
[INFO] Loading definitions file: @roblox - secondlife.d.luau
```

generated/secondlife.d.luau
```diff
@@ -466,31 +466,33 @@ declare quaternion: ((x: number, y: number, z: number, s: number) -> quaternion)
   toup: (q: quaternion) -> vector,
 }
 
+--[[ commented out to avoid shadowing magic functions find, format, gmatch, and match
```
tapple: If it were possible, I'd also comment out the `table` module from `secondlife.d.luau` to avoid shadowing the magic type functions `pack`, `freeze`, and `clone`. But, that is not possible due to the addition of `shrink`, `append`, and `extend`

slua_definitions.schema.json
tapple: if you like the dict, I'd like to do something similar with all the `-semantics` flags, especially since I added 3 more in #113

generated/secondlife.d.luau
HaroldCindy: That's fine, I imagine we can special-case inside the definition loader to remove things where we want to rely on the built-in logic? Or what do you think.

tapple: that would involve distributing a custom luau-lsp, which I don't think is worth it. Not with the viewer being much lower-hanging fruit where we can customize `EmbeddedBuiltinDefinitions.cpp` directly, and not worry about definition files.

Still, it might be worth bringing up with lute, once it supports definition files at all. Selene has this feature: standard library files are additive, even inside library tables, unlike luau-lsp definitions, where they replace

HaroldCindy (approved):
This seems reasonable to me, thanks! I'd really like for there to be some sort of way to "mix in" additions to `table`, `string`, and `bit32`, but that can be dealt with later.

PR #133, opened 2026-05-13
Merge `builtin-functions` into `global-functions`

tapple:
- Part of #122 
- `global-functions` are global functions that slua added
- `builtin-functions` are global functions that came with luau

There's really no reason to separate them. Combine them into one yaml section. I also properly sorted the global slua functions. They weren't before.

I made sure nothing changed by
1. Sorting the selene/keywords global functions together in the first commit, without modifying `slua_definitions.yaml`
2. Verifying that nothing changed in those files in the following commits

PR #135, opened 2026-05-27
Fix Luau global function type syntax errors

tapple:
- Follows #133 
- Part of #122 

Fix a number of issues with the global funcion types that were added to `secondlife.d.luau` in #133:
- They had syntax errors, especially with type packs
- They did not match `EmbeddedBuiltinDefinitions.cpp`

### DIFF
--- a/generated/secondlife.d.luau
+++ b/generated/secondlife.d.luau
@@ -251,7 +251,7 @@ declare function require(target: any): any
 declare function select<A...>(i: string | number, ...: A...): ...any
 declare setfenv: nil
 declare function setmetatable<T, MT>(t: T, mt: MT): setmetatable<T, MT>
-declare function tonumber(value: string, base: number?): number?
+declare function tonumber(value: string? | number, base: number?): number?
 declare function toquaternion(value: string? | quaternion): quaternion?
 declare function torotation(value: string? | quaternion): quaternion?
 declare function tostring<T>(value: T): string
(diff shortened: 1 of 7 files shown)


generated/secondlife.d.luau
```diff
@@ -230,37 +230,37 @@ export type PrimParamsSetterType = typeof(
...
+declare function require(target: any): any
+declare function select<A...>(i: string | number, ...: A...): ...any
 declare setfenv: nil
-declare function setmetatable(t: { [any]: any }, mt: { [any]: any }?): ()
-declare function tonumber(s: string, base: number?): number?
-declare function toquaternion(val: string? | quaternion): quaternion?
-declare function torotation(val: string? | quaternion): quaternion?
-declare function tostring(obj: any): string
+declare function setmetatable<T, MT>(t: T, mt: MT): setmetatable<T, MT>
+declare function tonumber(value: string, base: number?): number?
```
HaroldCindy: Doesn't `tonumber()` take `(string | number)?` rather than `string`?

HaroldCindy (changes requested):
@tapple looks good! Just one comment about `tonumber()` (and perhaps other casting functions) then this should be good to go.

generated/secondlife.d.luau
tapple: fixed: https://github.com/secondlife/lsl-definitions/pull/135/changes/242ac29597ba8b9f663c9433f262fbe72d6a3bfd

HaroldCindy (approved):

PR #137, opened 2026-06-02
Changing arguments names and improving tooltips

Suzanna-Linn:
This PR does two different but related things:
- Changes the names of arguments in functions and events to their Wiki names.
- Completes and improves the tooltips for functions, events, their arguments, and constants.

I originally wanted to make two separate PRs, but some of the tooltips refer to the argument names, so they have to go together. 
Everything was generated by script, so there shouldn't be any manual typos, only potential scripting errors.

Why change the argument names to match the Wiki?
- Avoids keyword conflicts: Some current Yaml arguments are named Key, Vector, or Rotation. If we lowercase them directly, they become reserved keywords in LSL/Lua and break. Using the Wiki names avoids this.
- Standardization: Yaml names currently start with an uppercase letter, which isn't standard in Lua.
- Accuracy: Wiki names have been heavily checked and corrected by the community over time.
- Conciseness: Wiki names are often shorter, whereas some of the current Yaml names are too verbose.
- Documentation consistency: The Wiki texts already use these argument names (Although hey could be changed, since usually they are in bold).

I have published the intermediate results in JSON format to help review these changes:
- All Yaml and Wiki names: https://suzanna-linn.github.io/slua/wiki-args-names.html
- Only names with changes (not including lettercase changes): https://suzanna-linn.github.io/slua/wiki-args-names-changes.html
- Only names with significant changes (not including changes like Position - pos, Value - val, etc): https://suzanna-linn.github.io/slua/wiki-args-names-big-changes.html

For functions that don't exist on the Wiki, I kept their original argument names but changed them to lowercase:
- llClearExperience
- llClearExperiencePermissions
- llFindNotecardTextCount
- llGetExperienceList
- llSetExperienceKey

Tooltip Improvements
- This PR adds and improves tooltips for functions, events, constants, and arguments based on data scraped from the Wiki.
- This part is still incomplete. While there are a lot of new and improved tooltips, these are just what could be scraped from the Wiki. We will need a second phase later to manually add the remaining ones.

Intermediate results in JSON format, showing where the new tooltips come from:
- functions and events, and their arguments: https://suzanna-linn.github.io/slua/wiki-tooltips-functions-events.html"
  - "tooltip" : the current tooltip
  - "description" : the description on Wiki
  - "newtooltip" : the new tooltip obtained from the previous ones
  - the argument key is the current Yaml name, "newname" is the new name in this PR.
- constants: https://suzanna-linn.github.io/slua/wiki-tooltips-constants.html
[...]

PR #140, opened 2026-06-03
Deprecate unused permissions

tapple:
None of these permissions are used by any function. Mark them as deprecated:
- PERMISSION_REMAP_CONTROLS
- PERMISSION_RELEASE_OWNERSHIP
- PERMISSION_CHANGE_JOINTS
- PERMISSION_CHANGE_PERMISSIONS

Noticed them while reviewing #137. Also includes the description case fix from that PR, to reduce merge conflicts


generated/experimental/enums.txt
```diff
@@ -1080,13 +1080,13 @@ enum RezParam {
 enum ScriptPermission {
   DEBIT = 0x2
   TAKE_CONTROLS = 0x4
-  REMAP_CONTROLS = 0x8
+  REMAP_CONTROLS = 0x8 // deprecated
```
tapple, line 1083: of the 4 permissions, `PERMISSION_REMAP_CONTROLS` seems the most likely to be useful soon, with work being done on `game_control`

HaroldCindy:
@AndrewMeadows Are you using any of these for `game_control` or do you think these can be marked deprecated?

AndrewMeadows:
I'm not using any runtime permissions for **game-control** yet. I'm fine with marking unused permissions as deprecated: perhaps that will make them available for repurposing in the future. Full speed ahead.

I don't think we'll want to allow scripts to remap game-control inputs. The current plan is to provide alternate, semantic names for `AXIS_0` - `AXIS_5` inputs (e.g. something like `AXIS_FWD`, `AXIS_STRAFE`, `AXIS_YAW`, `AXIS_PITCH`, `AXIS_UP`) and we won't want to scramble those meanings.

HaroldCindy (approved):

tapple:
My customers will definitely want to customize how the 20 or 30 game_control buttons I'll be using map to their keyboard. or they'll just continue using 30 gestures mapped to keyboard buttons. I dunno. gestures are fine if I don't need the key-up event

HaroldCindy:
Yeah, though I imagine that wouldn't require a _script_ to be allowed to remap those binds. At _most_ I'd expect a script using `game_controller` to expose some kind of control set name / type so those keybinds could be configured per-use by the viewer.

Seems like `PERMISSION_REMAP_CONTROLS` isn't really necessary then and could be deprecated.

PR #141, opened 2026-06-03
Deprecate unimplemented `PRIM_MEDIA_ALT_IMAGE_ENABLE`

tapple:
[According to the wiki](https://wiki.secondlife.com/wiki/LlSetPrimMediaParams), these prim media parameters are not implemented. Mark them as deprecated:
- PRIM_MEDIA_ALT_IMAGE_ENABLE

I noticed them while reviewing #137. Also includes the description improvement from that PR, to reduce merge conflicts


generated/experimental/enums.txt
```diff
@@ -889,15 +889,15 @@ enum PrimMediaError {
   WHITELIST_FAILED = 2001
 };
 enum PrimMediaParam {
-  ALT_IMAGE_ENABLE = 0
+  ALT_IMAGE_ENABLE = 0 // deprecated
```
tapple, line 894: Can confirm that these don't seem to do anything:
```lsl
default {
    state_entry() {
        print(llSetPrimMediaParams(0, [
            PRIM_MEDIA_HOME_URL, "https://secondlife.com",
            PRIM_MEDIA_FIRST_CLICK_INTERACT, TRUE,
            PRIM_MEDIA_ALT_IMAGE_ENABLE, TRUE
        ]));
    }

    touch_start(integer count) {
        print("touched");
    }
}
```

- `touch_start` never triggers on the media face, either before or after the media is started
- when media is not started, it shows plywood cube or whatever other texture I set, not some alt image

HaroldCindy:
Hmmm, not super sure about this one. It looks like the server does actually respect them, and will send them over if set, but the viewer may or may not use them properly.

I'm not sure if other viewers haven't been changed to actually respect these flags though.

tapple:
How do these compare to `PRIM_CAST_SHADOWS` and `STATUS_CAST_SHADOWS` in terms of actually making it to the viewer?

lsl_definitions.yaml
```diff
@@ -3625,9 +3625,11 @@ constants:
     value: 3
   PRIM_MEDIA_ALT_IMAGE_ENABLE:
```
tapple, line 3636: Getting out my crystal ball, `PRIM_MEDIA_ALT_IMAGE_ENABLE` seems to have an extremely low chance of ever being implemented, unless I grossly misunderstand it:

The intention seems to change one of these. I'm not sure which:
1. What does the viewer show when the media is not playing? currently, the regular blinn-phong or pbr texture
2. What does the viewer show after the user presses play but before the media downloads? currently, a black texture
3. What does the viewer show if the url is invalid? and it isn't a web page? Currently, whatever error response the server or chrome actually returns

If the intent is 1, then it seems useless. Users can already specify that with the edit tools.
If the intent is 2, that seems like it might be hard:
1. It should probably be a texture, not a boolean
2. It won't appear very often at all
3. It's probably pretty hard to get it to show up identically cross-platform
4. It may be hard to detect when the media has loaded
5. It may be hard to change this without changing the appearance of some media

If the intent is 3, then that also seems silly: displaying whatever the server actually returns seems more useful than trying to replace it based on some heuristics in the headers

lsl_definitions.yaml
```diff
@@ -3677,7 +3679,10 @@ constants:
     value: 2
   PRIM_MEDIA_FIRST_CLICK_INTERACT:
```
tapple, line 3690: Getting out my crystal ball, `PRIM_MEDIA_FIRST_CLICK_INTERACT` seems to have an unlikely chance of ever being implemented, but not as low as `PRIM_MEDIA_ALT_IMAGE_ENABLE`. Mostly, I'm just not sure what the intent or use case is:

What does interact mean? 
- `touch_start`? It can't send a browser event to the web page if the web page and it's javascript isn't loaded.  What would a script do with this touch? It could tell the user to make sure media is enabled, but, that's not super necessary as creators usually use the fallback texture for that, which doesn't require a click
- Does it "remember" the click, wait for the media to load (a nebulous concept already), and then replay the click, and hope it didn't click something they probably didn't mean to? that seems dangerous.

tapple: - There is a canny to enable this, with 2 votes: https://feedback.secondlife.com/feature-requests/p/enable-prim-media-first-click-interact-conditionally-in-the-viewer
- There is a merged pull request to enable this: https://github.com/secondlife/viewer/pull/4177

lsl_definitions.yaml
tapple: There are no open requests to enable this: https://feedback.secondlife.com/search?search=PRIM_MEDIA_ALT_IMAGE_ENABLE

lsl_definitions.yaml
tapple: @DarlCat can you enlighten me what this does?

tapple: Due to the fact that `PRIM_MEDIA_FIRST_CLICK_INTERACT` probably does something (still figuring out what), and the wiki is probably wrong about it, I reverted it's deprecation recommendation

tapple: I updated the wiki based on further testing, and hints from the PR: https://wiki.secondlife.com/w/index.php?title=Template:LSL_Constants/PrimMedia&diff=prev&oldid=1218877

Gets / Sets what happens when someone click on a playing, but unfocused media face:
* TRUE: Does the same thing as a focused media face does, and forwards the click to the web browser.
* FALSE: Focuses the media prim so that future clicks are forwarded to the web browser.
**Note:** This flag only works on HUDs. All other media faces always behave as if this flag is FALSE.

tapple: I updated the description in `lsl_definitions.yaml` in a different PR: #142 https://github.com/secondlife/lsl-definitions/pull/142/changes/51e448aafb53f08abba6fb68baf12c1460222fc4

tapple: After [speaking to Darl on discord](https://discord.com/channels/677442248157167619/1357059883400167585/1512242547777601637):
1. I figured out what `PRIM_MEDIA_FIRST_CLICK_INTERACT` is supposed to do, and [edited the wiki](https://wiki.secondlife.com/wiki/LlSetPrimMediaParams) accordingly (edited table entry and added Caveat)
2. I [filed a bug report](https://feedback.secondlife.com/bug-reports/p/prim-media-first-click-interact-settings-are-not-working-correctly) because it's currently misbehaving

lsl_definitions.yaml
tapple: To check what happens when a video url is correct vs incorrect, I used these 2 lua scripts. 

This one points to the correct URL, and plays the video on the secondlife.com home page (heavily featuring my product. yay!):
```luau
ll.SetPrimMediaParams(0, {
    PRIM_MEDIA_CURRENT_URL, 
"https://cdn.prod.website-files.com/60a83025d932b623a6a61ff8%2F67c73a654d7e0880e3668eeb_2025LandingPageVideo-transcode.mp4",
    PRIM_MEDIA_ALT_IMAGE_ENABLE, true,
    PRIM_MEDIA_AUTO_PLAY, true,
    PRIM_MEDIA_AUTO_LOOP, true,
})
```

This one has a typo in the url path and displays an xml error message:
```luau
ll.SetPrimMediaParams(0, {
    PRIM_MEDIA_CURRENT_URL, 
"https://cdn.prod.website-files.com/typo60a83025d932b623a6a61ff8%2F67c73a654d7e0880e3668eeb_2025LandingPageVideo-transcode.mp4",
    PRIM_MEDIA_ALT_IMAGE_ENABLE, true,
    PRIM_MEDIA_AUTO_PLAY, true,
    PRIM_MEDIA_AUTO_LOOP, true,
})
--]]
```

This one has a typo in the domain and displays the same error message that chrome would:
```luau
ll.SetPrimMediaParams(0, {
    PRIM_MEDIA_CURRENT_URL, 
"https://lol.website-files.com/60a83025d932b623a6a61ff8%2F67c73a654d7e0880e3668eeb_2025LandingPageVideo-transcode.mp4",
    PRIM_MEDIA_ALT_IMAGE_ENABLE, true,
    PRIM_MEDIA_AUTO_PLAY, true,
    PRIM_MEDIA_AUTO_LOOP, true,
})
```

lsl_definitions.yaml
tapple: Further testing revealed my bug is invalid. I closed it and repaired the parts of the wiki that referenced it

Martin-Pitt: The first click was actually relatively recently implemented and should have been merged into the viewer already I believe, not sure if firestorm has it yet or if it was implemented in firestorm instead. I just know that someone talked at length about having implemented it! Maybe Wolfgang? Ah it was H Nuzzle

DarlCat: Yep, FS has had that for the last two main releases IIRC. 😀

monty-linden:
The two candidates under discussion are:
* PRIM_MEDIA_ALT_IMAGE_ENABLE
* PRIM_MEDIA_FIRST_CLICK_INTERACT 

correct?

Those are active flags that are kept and participate in persistence, equality tests, etc..  The values are serialized in and out of the ObjectMedia Cap so are available for interpretation by any viewer that wants to do something with it.  Don't trust the wiki statements.

monty-linden (changes requested):
I believe both fields are live data and available as hints to viewers that would like to do something with them.  No need to deprecate.

PR #142, opened 2026-06-03
Improve tooltips for LSL constants (Wiki extract)

tapple:
I extracted just the constant tooltip changes from Suzanna's #137 as I believe those are ready for review, unlike the rest of the PR

The commits are as follows:
1. Suzanna's unmodified wiki extract from #137 
2. Revert all parts of it except the constants
3. Revert all parts of it that only change formatting
4. Tapple's changes start here

Since this is a wiki extract, the changes I made/will make in commits 4 and beyond should be backported to the wiki. My progress on that is:
- Not started


lsl_definitions.yaml
```diff
@@ -2758,2186 +2998,2462 @@ constants:
...
+      it).
     type: integer
     value: '0x02'
   PU_FAILURE_NO_NAVMESH:
     member-of: [CharacterPathUpdateType]
-    tooltip: This is a fatal error reported to a character when there is no navmesh
-      for the region. This usually indicates a server failure and users should file
-      a bug report and include the time and region in which they received this message.
+    tooltip: Path update fatal error triggered when there is no navmesh generated
+      for the region (typically indicates a server failure).
```
tapple, line 4900: I considered restoring the text about filing a bug report, but was unsure. Thoughts?

lsl_definitions.yaml
```diff
@@ -4948,893 +5464,982 @@ constants:
...
+    tooltip: Vehicle type constant for balloons that use hover and friction but no
+      deflection.
     type: integer
     value: 5
   VEHICLE_TYPE_BOAT:
     member-of: [VehicleType]
-    tooltip: Hovers over water with lots of friction and some anglar deflection.\nSee
-      http://wiki.secondlife.com/wiki/VEHICLE_TYPE_BOAT
+    tooltip: Vehicle type constant for boats that hover over water with high
+      friction and minor angular deflection.
```
tapple, line 6434: I'm unsure if removing the wiki link here and in the other `VEHICLE_TYPE_`s is a good idea or not. It shows the equivilancy of each vehicle type to the vehicle parameters

tapple:
- Closing; this is now merged into #144

PR #143, opened 2026-06-04
Rulesets for table-based function params

Rider-Linden:
Generates a lua interface for functions requiring structured lists of parameters with an initial implementation for `llprim.particleSystem`, `llprim.setMedia`, and `llhttp.request`

I selected each of those three because each had a subtly different catch.

### related PRs:
https://github.com/secondlife/slua/pull/86
https://github.com/secondlife/server/pull/2574

```lua
llprim.particleSystem({
    pattern       = PSYS_SRC_PATTERN_ANGLE_CONE,
    texture       = "da8e96f5-4ada-4f37-bd2c-cf3e68c49a42",
    color_begin   = vector(1, 0.6, 0.1),
    color_end     = vector(1, 0.1, 0),
    alpha_begin   = 1.0,
    alpha_end     = 0.0,
    scale_begin   = vector(0.1, 0.1, 0),
    scale_end     = vector(0.4, 0.4, 0),
    glow_begin    = 0.5,
    glow_end      = 0.0,
    accel         = vector(0, 0, 0.5),
    burst_speed_min = 0.5,
    burst_speed_max = 1.5,
    burst_rate    = 0.05,
    burst_count   = 5,
    burst_radius  = 0.1,
    src_max_age   = 0,     -- 0 = continuous
    part_max_age  = 3.0,
    color_interp  = true,
    scale_interp  = true,
    emissive      = true,
    wind          = false,
}, link)

-- note link is an optional parameter.  If missing it defaults to LINK_THIS
```

`setMedia` ends up looking like this:
```lua

local link = nil

llprim.setMedia(0, {
    current_url          = "https://example.com/",
    home_url             = "https://example.com/",
    auto_play            = true,
    auto_loop            = false,
    controls             = PRIM_MEDIA_CONTROLS_STANDARD,
    width_pixels         = 1024,
    height_pixels        = 768,
    whitelist_enable     = true,
    whitelist            = { "example.com", "*.example.com" },
    perms_interact       = PRIM_MEDIA_PERM_ANYONE,
    perms_control        = PRIM_MEDIA_PERM_OWNER,
}, link)

-- Note, the first parameter is the face.  The last parameter is the link number.  If link number is omitted 
-- it defaults to LINK_THIS
```
http request:
```lua
local request_id = llhttp.request(
    "https://api.example.com/data",
    {
        method          = "POST",
        mimetype        = "application/json",
        verify_cert     = true,
        extended_error  = true,
        custom_header   = {
            ["X-Api-Key"]        = "my-secret-key",
            ["X-Request-Source"] = "second-life",
        },
        accept          = { "application/json", "text/plain" },
    },
    '{"hello": "world"}'
)
```


lsl_definitions.yaml
```diff
@@ -4235,56 +4247,66 @@ constants:
...
       \ NOT have particles in it."
     type: integer
     value: 10
+    value-type: float
   PSYS_SRC_MAX_AGE:
     member-of: [ParticleParam]
     tooltip: How long this particle system should last, 0.0 means forever.
     type: integer
     value: 19
+    value-type: float
```
tapple, line 4830: yes. this is where I was hoping you'd put this data, rather than in the ruleset section

lsl_definitions/slua.py
```diff
@@ -10,9 +10,23 @@
...
 from lsl_definitions.utils import Deprecated, remove_worthless
 
+_TABLE_RULESET_TYPE_MAP: dict[str, str] = {
+    "float": "number",
+    "integer": "number",
+    "vector": "vector",
+    "string": "string",
+    "key": "(string | uuid)",
+    "asset": "(string | uuid)",
+}
```
tapple, line 33: I'll formalize this better in a future PR. I do like this structure over all the loose `-semantics` fields laying around

lsl_definitions/generators/lscript.py
```diff
@@ -883,6 +884,111 @@ def gen_lua_constant_definitions(definitions: LSLDefinitions) -> str:
...
+    sections = []
+    for ruleset_name, ruleset_data in definitions.builder_rulesets.items():
+        if ruleset_data.get("type", "builder") != "table":
+            continue
+
+        enum_name = ruleset_data["enum"]
+        lua_module = ruleset_data["lua-module"]
+        lua_type = ruleset_data["lua-type"]
+        apply_fn = ruleset_data["apply-fn"]
+        apply_link_fn = ruleset_data["apply-link-fn"]
```
tapple: this will definitely need improvement at some point (now, later, whatever)                                                    
                                                                                                                              
apply_fn has the right signature for                                                                                                        
- llCreateCharacter                                                                                                           
- llParcelMediaCommandList                                                                                                    
- llSetCameraParams                                                                                                           
- llSetGroundTexture                                                                                                          
- llUpdateCharacter                                                                                                           
                                                                                                                              
apply_link_fn doesn't have the right signature for any function except for `llLinkParticleSystem`, but it's close for these (it's missing `face`):
- llSetLinkGLTFOverrides
- llSetLinkMedia
- llSetPrimMediaParams
                                                                                                                              
neither one has the right signature for                                                                                                     
- llRezObjectWithParams                                                                                                       
- llSetAgentEnvironment                                                                                                       
- llSetEnvironment                                                                                                            
- llCastRay                                                                                                                   
- llEvade (no params yet)                                                                                                     
- llFindNotecardTextCount (no params yet)                                                                                     
- llFindNotecardTextSync (no params yet)                                                                                      
- llFleeFrom (no params yet)                                                                                                  
- llGetAgentList (no params yet)
- llGetAttachListFiltered                                                                                                     
- llGetClosestNavPoint                                                                                                        
- llGetStaticPath                                                                                                             
- llGiveAgentInventory                                                                                                        
- llHTTPRequest                                                                                                               
- llMapBeacon                                                                                                                 
- llNavigateTo                                                                                                                
- llPatrolPoints                                                                                                              
- llPursue                                                                                                                    
- llSetKeyframeMotion                                                                                                         
- llSetParcelForSale                                                                                                          
- llTransferOwnership (no params yet)                                                                                         
- llWanderWithin

generated/secondlife_selene.yml
```diff
@@ -15696,6 +15696,137 @@ structs:
...
+      property: full-write
+      type: number
+    accel:
+      property: full-write
+      type:
+        display: vector
+    pattern:
+      property: full-write
+      type: number
+    innerangle:
```
tapple: Alias suggestions. I wanted them all to be consistently `subject_adjective`. I also changed `start` to `begin` for consistency with `angle_begin`, which I think is already in the right form
- `innerangle` -> `angle_inner`
- `outerangle` -> `angle_outer`
- `start_scale` -> `scale_begin`
- `end_scale` -> `scale_end`
- `start_color` -> `color_begin`
- `end_color` -> `color_end`
- `start_alpha` -> `alpha_begin`
- `end_alpha` -> `alpha_end`
- `start_glow` -> `glow_begin`
- `end_glow` -> `glow_end`
- `interp_color` -> `color_interp`
- `interp_scale` -> `scale_interp`

tapple (commented):
Alias suggestions. I wanted them all to be consistently `subject_adjective`. I also changed `start` to `begin` for consistency with `angle_begin`, which I think is already in the right form

generated/secondlife_selene.yml
tapple: Alias suggestions. I wanted them all to be consistently `subject_adjective`. I also changed `start` to `begin` for consistency with `angle_begin`, which I think is already in the right form
- `innerangle` -> `angle_inner`
- `outerangle` -> `angle_outer`
- `start_scale` -> `scale_begin`
- `end_scale` -> `scale_end`
- `start_color` -> `color_begin`
- `end_color` -> `color_end`
- `start_alpha` -> `alpha_begin`
- `end_alpha` -> `alpha_end`
- `start_glow` -> `glow_begin`
- `end_glow` -> `glow_end`
- `interp_color` -> `color_interp`
- `interp_scale` -> `scale_interp`

generated/cpp/fluent_builder_descriptors.cpp
```diff
@@ -0,0 +1,47 @@
+// particle-params
+static const FluentParamDescriptor kParticleParamsDescs[] = {
+    {"flags", 'i', 0},
```
tapple, line 31: if both flags and booleans are given, they are `or`ed together?

lsl_definitions.yaml
```diff
@@ -16105,3 +16133,15 @@ builder-rulesets:
...
+  particle-params:
+    type: table
+    enum: ParticleParam
+    filler-tokens: [part, src]
+    flag-enum: ParticleFlag
+    flag-field: flags
+    flag-mask: _MASK
+    lua-module: llparticle
+    lua-type: ParticleParams
+    apply-fn: ParticleSystem
```
tapple: I don't think `llParticleSystem` needs a lua wrapper. It's covered by a `llLinkParticleSystem` wrapper that defaults to `LINK_THIS`. that's what `ParamSetter` does, I think (but there it's also the only 0-delay variant)

slua_definitions.yaml
```diff
@@ -318,6 +318,27 @@ classes:
...
+  properties: []  # injected at generation time from expand_table_ruleset("particle-params")
+  functions:
+  - name: new
+    comment: Create a new ParticleParams, optionally bulk-initializing from a table.
+    parameters:
+    - name: init
+      type: '{[string]: any}?'
+    return-type: ParticleParams
+  methods:
+  - name: apply
```
tapple: from Harold:
Is the builder itself necessary there? What utility does it provide over an exported declared type for the dict-like table and a function that accepts tables of that shape? That's generally how similar APIs work in other Luau projects, the "builder-ness" of SPP isn't something that's really needed elsewhere.
That way serialization is easier as well, since you don't need to preserve the type

I don't know if a fluent wrapper is really what we want / need here. Just declaring an expected table shape and a function that consumes it should be fine, since that's really all the builder does

tapple: yeah. that's true. most lua api's would be like 
```luau
llprim.particleSystem{
    emissive = true,
    target_linear = true,
    pattern = PSYS_SRC_PATTERN_EXPLODE,
    start_color = vector(0, 0, 1),
    src_max_age = 10,
    burst_rate = 0.1,
    burst_count = 10,
    target_key = ll.GetOwner()
}
```
"just a function" would work better for most of the param-dict wrappers: https://github.com/secondlife/lsl-definitions/pull/143#discussion_r3358556418

tapple: I'm unsure whether those functions should look like 1. 
```luau
function llprim.setMedia(link: number, face: number, params: {})
```
to match the lsl function, or like 2. 
```luau
function llprim.setMedia(params: {}, link: number?, face: number?)
```
so the link/face can be optional (defaulting to `LINK_THIS` and `ALL_SIDES`, or 
3. mix the two and make a mess like `table.insert` does:
```luau
llprim.setMedia: ((link: number, face: number, params: {}) -> ())
    & ((face: number, params: {}) -> ())
    & ((params: {}) -> ())
```
or, 4. just add link number/face number as optional dict params, and leave positional params out of it entirely

tapple: oh. the dict parameter should be optional too, and work the same as an empty list. Most of the param-dict functions make sense with an empty param list

lsl_definitions/generators/lscript.py
```diff
@@ -883,6 +884,111 @@ def gen_lua_constant_definitions(definitions: LSLDefinitions) -> str:
...
+        desc_lines = []
+        for desc in expand_table_ruleset(definitions, ruleset_name):
+            name = desc.pretty_name if desc.pretty_name else desc.strict_name
+            sem = _semantic_map[desc.value_type]
+            desc_lines.append(f"    {{\"{name}\", '{sem}', {desc.tag}}},")
+
+        body = "\n".join(desc_lines)
+        section = (
+            f"// {ruleset_name}\n"
+            f"static const FluentParamDescriptor {array_name}[] = {{\n{body}\n}};\n"
```
tapple: Is it still true that the bulk of the code to generate the builder lives in lsl-definitions, and slua/tools/gen_primparams_methods.py calls that code and then converts it to c struct format? just thinking ahead to what changing the structure of rulesets in the yaml file would actually look like given that it's growing in capability

tapple: from Rider:
I believe that yes, gen_primparams_methods is called out of the slua repo.
I'm not entirely happy with that design since it means that slua needs to match the simulator.

That implies that ANY change to lsl_definitions requires a recompile on the VM as well.
(spelling correction to a tooltip?  new VM)

tapple: I got the idea somewhere in my mind that there was an intermediate json file that gen_all_definitions wrote and that gen_primparams_methods read, but, I can't find anything like that in reality
would probably be better if that was reality, so that slua only needed the lsl-definitions build, not lsl-definitions itself
but, I can understand why it isn't that way right now. PrimParams isn't even in beta yet, unless I missed it

tapple: from Rider:
It has to do with autobuild and version numbers.  build lsl_definitions and the version goes up.  Update the simulator and the version no longer matches the one use to build the VM and autobuild screams about version mismatches.

tapple: from Harold:
prior feedback RE the codegen for those wrappers being in slua proper so particles can be mocked / re-implemented for local dev / testing as with the SPP wrapper. There's not really a better place to put them other than slua that keeps the local dev story sane. Ideally people should only have to re-implement the very low-level implementation details of the thing that our wrappers wrap (ll.ParticleSystem here, since that's the thing that does the work) so things like do local simulation a la LSLForge can work correctly without a ton of work on their part.

lsl_definitions.yaml
```diff
@@ -4203,24 +4209,28 @@ constants:
     value: '0x400'
   PSYS_PART_START_ALPHA:
     member-of: [ParticleParam]
+    pretty-name: alpha_begin
```
tapple, line 4710: nice. I like `pretty-name` better than `method-name` used in rulesets

generated/secondlife.d.luau
```diff
@@ -432,6 +424,7 @@ declare lljson: {
 
 declare llprim: {
   ParamsSetter: PrimParamsSetterTypeMeta,
+  ParticleSystem: (params: ParticleParams?, link: number?) -> (),
```
tapple: shouldn't this be `particleSystem` or `particlesystem`, not `ParticleSystem`, since it's a function, not a class?

unsure, because we have conflicting styles:
- in `ll` module, the convention would say `ParticleSystem` (compare `ParticleSystem`)
- in `PrimParams` class, the convention would be `particleSystem` (compare `physicsMaterial`)
- in `table` module, the convention would be `particlesystem` (compare `isfrozen`)

Rider-Linden: Yes, it should.  I thought I had switched it back. 
I'll update the PR description later, 

I also want to add a second class. llSetLinkMedia, I'm going back and forth about whether or not do try something fancy wit PRIM_MEDIA_WHITELIST, but will probably save that for a second pass.

slua_definitions.yaml
Rider-Linden: nil and an empty table should both behave like an empty list.

generated/secondlife.d.luau
Rider-Linden: When I rewrite the PR I'll call this out above but scripts end up looking like this:

```lua
local def: ParticleParams = {
    emissive = true,
    target_linear = true,
    pattern = PSYS_SRC_PATTERN_EXPLODE,
    color_begin = vector(0, 0, 1),
    src_max_age = 10,
    burst_rate = 0.1,
    burst_count = 10,
    target_key = ll.GetOwner()
}

local def2: ParticleParams = {
    color_interp = true,
    emissive = true,
    pattern = PSYS_SRC_PATTERN_EXPLODE,
    color_begin = vector(1, 0, 0),
    color_end = vector(0, 1, 0),
    alpha_begin = 1.0,
    alpha_end = 1.0,
    src_max_age = 5.0,
    burst_rate = 0.1,
    burst_count = 10,
    burst_radius = 2
}

local function do_touch(detected)
    llprim.ParticleSystem(def)
    
    llprim.ParticleSystem(def2,  2)
end

LLEvents:on("touch_start", do_touch)

print(lljson.encode(def))
print(lljson.encode(def2))
```

generated/secondlife.d.luau
```diff
@@ -85,6 +85,74 @@ type LLJsonDecodeOptionsWithPath = {
   reviver: (key: string | number, value: any, parent: {}?, ctx: {path: {string | number}}) -> any
 }
 type LLJsonDecodeOptions = LLJsonDecodeOptionsWithoutPath | LLJsonDecodeOptionsWithPath
+export type ParticleParams = {
+  flags: number?,
+  color_begin: vector?,
```
gwigz, line 90: reckon it's possible to document the defaults some of these values have?

Rider-Linden: @Suzanna-Linn Has a PR with an RFC proposing that: https://github.com/secondlife/lsl-definitions/pull/160

monty-linden (approved):
zoooom

PR #149, opened 2026-06-14
Adding parameters and types to the metamethods

Suzanna-Linn:
Adding parameters and types to the metamethods:
- structured like the functions.
- "tooltip" changed to "comment" as in the functions.


slua_definitions.yaml
```diff
@@ -2880,46 +2880,222 @@ builtin-constants:
     value: 'false'
 metamethods:
   __add:
```
tapple: change this for consistency with the rest of the file:

```suggestion
  - name: __add
```

This will require python changes

slua_definitions.yaml
```diff
@@ -2880,46 +2880,222 @@ builtin-constants:
...
+        type: any
+        comment: "The left operand of the multiplication."
+      - name: right
+        type: any
+        comment: "The right operand of the multiplication."
+    return-type: any
   __newindex:
-    tooltip: "Method or table called when table is indexed with an assignment."
+    comment: "Called when assigning a value to a key that does not exist in the table."
+    type: "(self: table, key: any, value: any) -> ()"
```
tapple: Is the overload field necessary? I could express the same thing without the overload just using comment and type:
```suggestion
    comment: Called when assigning a value to a key that does not exist in the table.
        Can also be a fallback dictionary/table where new key-value assignments are forwarded.
    type: "((self: table, key: any, value: any) -> ()) | {}"
```

tapple: overload is also not an accurate name for the field. Unsure what I'd want to call it if it needed to stay. I'd prefer it was gone. The `type` field you added vs functions already does everything that function overloads currently do in the generators. It might be a better solution than overloads everywhere in the file.

slua_definitions.yaml
Suzanna-Linn: I was going to do it, but I thought that it would break the generator, so I kept the current format.

If you update the generator (I don't know python), I will change the yaml.

slua_definitions.yaml
Suzanna-Linn: It's important for documentation (*) to be able to generate two different signatures for `__index` and `__newindex`, since the way to use them is very different.
These metamethods are difficult to learn for SLua beginners and it's good to be as clear as possible.
Displaying two different signatures is more expressive than explaining it in the comment.

Indeed `overload` is not the best way, it's just reusing what is already there.

I suggest:
- defined in one piece as `function | table` and no `overload`
- a new field `signatures`with the two signatures

This will be useful for other functions, `math.random()`, for instance, that has three signatures.

(*) in theory, I don't know Linden plans about documentation...

tapple: I currently have very little faith that any documentation will get made that we don't make, so, if you have a plan for metamethod documentation, just do that. I changed rider's mind about how plugins can work by just doing the work to make slua_definitions.yaml. If you do the work, the lindens will probably use it. There's more of us than there are lindens

slua_definitions.yaml
tapple: ok. I'll do that this week sometime

Suzanna-Linn: I have changed metamethods from object to array with the field "name:".

And I have done a change in the generator to make it work some way. Please look at it to do it well.

slua_definitions.yaml
Suzanna-Linn: I have changed `__index` and `__newindex`.

Now they have only one definition with a union type, no `overload:`.

I have added a new field `variants:` for the different signatures.

The idea is:
- code generators will ignore `variants:`.
- documentation generators will use `variants:` only, when it is there.

If this is merged I will look for functions that could have variants.

slua_definitions.yaml
```diff
@@ -2879,47 +2879,246 @@ builtin-constants:
...
+          - name: key
+            type: any
+            comment: "The key that was not found in the table."
+        return-type: any
+      - comment: "A fallback dictionary/table used to search for missing keys."
+        type: "{ [any]: any }"
+        return-type: any
+  - name: __iter
+    comment: "Called before generalized iteration begins on a table or userdata. Returns an iterator generator function, a state object, and an initial control variable."
+    type-parameters: [S, K, V...]
```
tapple: There's trailing whitespace on this line. Run this to autofix this and one other occurance:
```
pip install pre-commit
pre-commit run --all
```

slua_definitions.yaml
tapple: Your change looks fine. I'll wait to make a validator until any generator actually uses the metamethod data, rather than just names. Maybe keywords.xml could use a tooltip. I dunno.

slua_definitions.yaml
Suzanna-Linn: Done.

slua_definitions.yaml
tapple: If you followed the example of the luau.org docs, your two ideas for `variants` and `type-doc` would do the same thing: https://luau.org/library/#debug-library

unsure if variants even belongs in the definitions file, rather than the article file for that function/category/library/whatever. Hard to say without any concrete way to combine definitions and articles yet

tapple: Anyway, variants seems fine for metamethods for now, mostly because metamethods don't have any generators. I'd be a lot more hesitant to add that to functions

Suzanna-Linn: It's more solid to have the variants in the same file, since they have to be always coordinated with any change in the function (like adding a new parameter), rather than having it in two different sources, probably updated by different people.

I'm not following luau.ord docs, just thinking on how I would do it, as a replacement for the how it will be really be done, which we don't know.

About my idea to add new fields "docs-type" and "docs-return-type" with more beginner friendly types, I will not do it, because it's possible to generate them by script.
I added a  "simple types" (at top right) in: https://suzanna-linn.github.io/slua/moving-lua-definitions

Suzanna-Linn: My idea is too add `variants:` to a few functions too, if this PR is approved.

Generators should just ignore the `variants:` field. It's for documentation generators only.

slua_definitions.yaml
```diff
@@ -2881,47 +2881,246 @@ builtin-constants:
...
-  __iter:
-    tooltip: "Method called to iterate over the table."
-  __idiv:
-    tooltip: "Overload method for // operator."
-  __tojson:
-    tooltip: "Called by lljson encode and slencode to encode the value as JSON."
-  __jsonhint:
-    tooltip: 'String property used by lljson.encode. Must be one of "array" or "object".'
+  - name: __add
+    comment: "Defines behaviour for the addition (+) operator."
```
monty-linden: For our docs and tooltips, "US English" is the default spelling style so "behavior" rather than "behaviour".  That occurs a few times and I haven't checked for other Britishisms but that's the one change needed.

monty-linden (changes requested):
Britishisms

slua_definitions.yaml
Suzanna-Linn: I have changed 13 "behaviour" and 3 "undergoing". I think that there aren't any more.

monty-linden (approved):

PR #150, opened 2026-06-15
Adding default values for the optional parameters

Suzanna-Linn:
Adding default values for the optional parameters with a new field `default-value:`.

I have updated the generator to make it work (although it hasn't generated changes), please check.


slua_definitions.yaml
```diff
@@ -1390,6 +1408,7 @@ modules:
       type: any
     - name: options
       type: LLJsonEncodeOptions?
+      default-value: nil
```
tapple:
```suggestion
      default-value: '{allow_sparse=false, skip_tojson=false, replacer=nil}'
```

slua_definitions.yaml
```diff
@@ -1410,6 +1430,7 @@ modules:
...
```
tapple:
```suggestion
      default-value: '{allow_sparse=false, skip_tojson=false, tight=false, replacer=nil}'
```

slua_definitions.yaml
```diff
@@ -1400,6 +1419,7 @@ modules:
       type: string
     - name: options
       type: LLJsonDecodeOptions?
+      default-value: nil
```
tapple:
```suggestion
      default-value: '{track_path=false, reviver=nil}'
```

slua_definitions.yaml
```diff
@@ -1420,6 +1441,7 @@ modules:
...
```
tapple:
```suggestion
      default-value: '{track_path=false, reviver=nil}'
```

slua_definitions.yaml
```diff
@@ -1738,9 +1763,11 @@ modules:
     - name: min
       comment: Minimum value of the range (optional).
       type: number?
+      default-value: 1
```
tapple:
```suggestion
      default-value: nil
```

`math.random()` and `math.random(1)` are different. `math.random` would be a good candidate for a function with a variant, as it returns integers with arguments and floats without. very weird.

slua_definitions.yaml
tapple, line 2378: `table.insert` should have a default for the middle parameter of '#a'

slua_definitions.yaml
```diff
@@ -2689,6 +2746,7 @@ modules:
     - name: axis
       comment: Axis to determine the sign of the angle (optional).
       type: vector?
+      default-value: nil
```
tapple, line 2751:
```suggestion
      default-value: 'vector.cross(a, b)'
```

slua_definitions.yaml
tapple, line 1934: `os.time` default value is `os.date("!*t")`

tapple (approved):
looks good. A few suggested improvements

slua_definitions.yaml
Suzanna-Linn: Done

slua_definitions.yaml
Suzanna-Linn: Done

slua_definitions.yaml
Suzanna-Linn: Done

slua_definitions.yaml
Suzanna-Linn: Done

slua_definitions.yaml
Suzanna-Linn: Done

Yes, `math.random()` is to use `variants:`.

slua_definitions.yaml
Suzanna-Linn: In this signature is not optional (`type: number`). It would look ugly to show a default value for a parameter without `? `.
In this case we will show the two signatures.

slua_definitions.yaml
tapple: I checked this by: 1. reading the source code, and 2. with this verification script:
```luau
local function randomVector()
    return vector.create(math.random()-0.5, math.random()-0.5, math.random()-0.5)
end

for i = 1, 1000 do
    local a = randomVector()
    local b = randomVector()
    local cross = vector.cross(a, b)
    local angle1 = vector.angle(a, b)
    local angle2 = vector.angle(a, b, cross)
    if angle1 ~= angle2 then
        print("a", a, "b", b)
    end
end
    
print("done")
```

Suzanna-Linn: I don't like it:
- It's confusing for non-math oriented scripters (like me).
- It suggests that it's better to pass `nil` explicitely to avoid the function calculating an extra cross vector.
- Internally the function checks for `nil`, it doesn't assign it a value.

slua_definitions.yaml
Suzanna-Linn: I don't like it:
- It suggests that `os.time()` is calculating a date, which would take extra time.
- Internally the function does a different process to get the current time.

This would be another function for `variant:`

slua_definitions.yaml
Suzanna-Linn: Yes, it's mathematically the same, but `vector.angle()` just checks `if (axis)`:

```js
static int vector_angle(lua_State* L)
{
    const float* a = luaL_checkvector(L, 1);
    const float* b = luaL_checkvector(L, 2);
    const float* axis = luaL_optvector(L, 3, nullptr);

    // cross(a, b)
    float cross[] = {a[1] * b[2] - a[2] * b[1], a[2] * b[0] - a[0] * b[2], a[0] * b[1] - a[1] * b[0]};

    double sinA = sqrt(cross[0] * cross[0] + cross[1] * cross[1] + cross[2] * cross[2]);
    double cosA = a[0] * b[0] + a[1] * b[1] + a[2] * b[2];
    double angle = atan2(sinA, cosA);

    if (axis)
    {
        if (cross[0] * axis[0] + cross[1] * axis[1] + cross[2] * axis[2] < 0.0f)
            angle = -angle;
    }

    lua_pushnumber(L, angle);
    return 1;
}
```

tapple: 1. You need to know math to use any of the vector functions. And, if you do know math, cross product is a better-defined thing than "angle". If it's documented that the default axis is the cross product, that tells me that, if axis is omitted, `vector.angle` always returns a positive number (This property should be documented explicitly though, for less math-inclined people. I noticed it's missing).
2. it's always better to leave out the optional parameter if the default is fine. That's true of all functions with a default parameter, and not unique to `math.angle`
1. all functions with an optional parameter check for nil. that's what makes them optional parameters. And, it does internally compute the cross product. The c code translates to lua like:

```luau
function vector.angle(a, b, axis)
    local cross = vector.cross(a, b)
    local angle = math.atan2(vector.magnitude(cross), vector.dot(a, b))
    if axis and vector.dot(cross, axis) < 0 then
        angle = -angle
    end
    return angle
end
```

slua_definitions.yaml
tapple: internally, none of the functions exactly replace the value with the documented default value. default value is a documentation fiction to illustrate better how the function works. My thoughts on default values is:
1. Does the function always act the same with the given expression substituted for the default value?
2. Does it help illustrate how the function works?

tapple: that works too.

slua_definitions.schema.json
```diff
@@ -114,6 +114,10 @@
                 "comment": {
                     "markdownDescription": "A brief description of this parameter.",
                     "type": "string"
+                },
+                "default-value": {
+                    "markdownDescription": "The default value of this parameter when omitted.",
```
monty-linden, line 119: A comment.  Reviewing changes in progress, we really are describing the 'effective value' of the parameter when omitted, not the literal 'default value'.

slua_definitions.yaml
```diff
@@ -673,9 +678,11 @@ functions:
   - name: i
     comment: Starting index (optional, defaults to 1).
     type: number?
+    default-value: 1
   - name: j
     comment: "Ending index (optional, defaults to #a)."
```
monty-linden: " (optional, defaults to #a)." no longer needed or correct.

slua_definitions.yaml
```diff
@@ -1149,6 +1158,7 @@ modules:
     - name: count
       comment: Number of bytes to write (optional).
```
monty-linden: " (optional)" not needed.  Or I'm assuming this information is now provided by the '?' suffix on type.  Maybe it's not the intent to have that be the only indicator.  Discuss.  (I won't mention any more occurrences.)

slua_definitions.yaml
monty-linden: I don't have an answer for this.  There are three options:
* Literal Value.  Typically going to be a 'nil' from Lua mechanisms and not interesting.
* Substituted Value.  The value provided by 'LuaL_optstring()' or whatever call when defaulted.  Direct substitution by scripter will give identical results.  Will sometimes also be boring like a 'nil'.
* Effective Value.  The value the captures the final behavior when defaulted.

The preference forming in my head is for the last for tooltips and the last two for longer docs (maybe just Effective Value when they're the same).

monty-linden: I'm thinking, let's take tapple's effective suggestions for tooltips and see how they land with residents.

slua_definitions.yaml
monty-linden: Let's take tapple's and see how it lands.  It **is** a case that might need a little more elaboration.  I don't like this specifically but this is the idea:
```
default-value: 'vector.cross(a, b) -- effective giving [0..pi] results'
```

monty-linden (changes requested):
One or two quick fixes in here.  Fixes/questions about the use of '(optional)' text in field description.  And willing to try out tapple's 'Effective Value' documentation for field defaults.  All open to debate.

slua_definitions.yaml
tapple: though, marking the middle argument as having a default would make it consistent with the upstream selene library files, which is why this is marked optional. The missing `?` is just because the overload describes it more accurately. If the documentation shows the overload (like luau.org) that's accurate too: https://luau.org/library/#table-library

tapple: also, selene is the generated file that harold and rider and wolf and I like the least, so, it shouldn't hold a lot of weight

tapple: Effective value is what I was thinking when I proposed these, though I didn't have that name. I was thinking "the most interesting value that gives the same result". Because, if it's an interesting value, it might give the reader insights about how other parts of the code relate to this function. Like:
- I understand angles, but, I'm unsure what axis is, and I don't know what a cross product is. What kind of axis does cross product produce, and why does it make the angle always positive?
- I understand "current time", but, can I get it in a different format? oh hey. `os.date` can do something. let's look at that

slua_definitions.yaml
tapple: as an aside, I learned vector math by messing around in SL. And, having experience with vector math scripting, cross product is pretty high on my vector function usage. Here's a rough estimate of how I'd rank vector functions by my own usage:
1. addition
2. quaternion rotate/multiply
3. ll.AxisAngle2Rot
4. vector.normalize
5. cross product
6. dot product
7. vector.magnitude

slua_definitions.yaml
tapple: this is mostly there for the tooltip generators (keywords.xml and secondlife.docs.json). The tooltip generator should be modified to add this data back in if it's removed from the raw text. This should be a follow-up PR, not this PR

tapple: Issue added: #159

slua_definitions.yaml
Suzanna-Linn: I changed #a to #tab in a previous PR, changing it again hoping to avoid unresolved conflicts.

About "optional", I'm answering in the other "optional" comment below.

slua_definitions.yaml
Suzanna-Linn: There are about 30 "optional"s, a very few of them with default value. I prefered to keep them to avoid messing with the tooltips in the editors.

I can remove them now, or in a next PR.

I'm planning to do a general review of the tooltips to expand some of them with more info. My idea was to remove the "optional"s then after coordinating work with the generators.

I don't know Python, probably I could make it work, probably not in a nice way, so better if Tapple volunteers to do it...

slua_definitions.yaml
Suzanna-Linn: Many scripters, including me, use vectors and rotations without knowing maths at all, based on code examples and snippets.

I'm horrified at the idea of having to explain this default value to students who don't know maths like me.

I'm ok with it if you like it, but please in another PR. I don't want to be "the one who added it" (because if it shows up in class, someone will ask).

slua_definitions.yaml
Suzanna-Linn: I decided to give it at try...
- "optional"s and "default"s removed.
- generator updated (please check).

monty-linden (approved):
Giving the updates a try.

PR #158, opened 2026-06-18
Fixing a few fields in metamethods

Suzanna-Linn:
Fixing a few fields in metamethods after reviewing again

### DIFF
--- a/slua_definitions.yaml
+++ b/slua_definitions.yaml
@@ -2895,10 +2895,10 @@ metamethods:
   - name: __call
     comment: "Called when the table or object is called as if it were a function."
     type-parameters: [A..., R...]
-    type: "<A..., R...>(self: table, A...) -> R..."
+    type: "<A..., R...>(self: {}, A...) -> R..."
     parameters:
       - name: self
-        type: table
+        type: '{}'
         comment: "The table being invoked."
       - name: ...
         type: "A..."
@@ -2950,10 +2950,10 @@ metamethods:
     return-type: any
   - name: __index
     comment: "Called when looking up a key that does not exist in the table."
-    type: "((self: table, key: any) -> any) | { [any]: any }"
+    type: "((self: {}, key: any) -> any) | { [any]: any }"
     parameters:
       - name: self
-        type: table
+        type: '{}'
         comment: "The table being accessed."
       - name: key
         type: any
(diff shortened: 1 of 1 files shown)


slua_definitions.yaml
```diff
@@ -2976,7 +2976,7 @@ metamethods:
   - name: __iter
     comment: "Called before generalized iteration begins on a table or userdata. Returns an iterator generator function, a state object, and an initial control variable."
     type-parameters: [S, K, V...]
-    type: "<S, K, V...>(self: any) -> ((S, K) -> (K?, V...), S, K)"
+    type: "<S, K, V...>(self: table) -> ((state: S, index: K) -> (K?, V...), S, K)"
```
tapple: won't the first parameter to all metamethods always be a table, not any? since only tables can have user-defined metatables

The best way to write table, at least in typechecker syntax, is:

```suggestion
    type: "<S, K, V...>(self: {}) -> ((state: S, index: K) -> (K?, V...), S, K)"
```

Since the typechecker will never see these, and they are only for documentation, `table` is fine too. Up to you which you think makes for better documentation. I know luau.org uses `table` not `{}`. Then again, they would also say `function` rather than `((state: S, index: K) -> (K?, V...), S, K)`. So, you're kinda mixing styles

tapple: if you want to add `type-doc` to metamethods, go ahead. Easier to get a feel for it here before adding it to all the functions

tapple: actually, luau.org says `<iterator>`: https://luau.org/library/#global-functions
```
function pairs(t: table): <iterator>
```
`((state: S, index: K) -> (K?, V...), S, K)` is indeed the precise form for `<iterator>`

Suzanna-Linn: I have changed `__len`, `__tostring` and `__unm` from 'any' to 'table'.

I think that with these ones they are all 'table'.

Not the binary operators, which first operand can be anything without `__add`, if the second operand has `__add`.

Suzanna-Linn: I don't think that it's important for documentation what style of type annotation is in the definitions, since it can be converted to a beginner friendly one when displaying it.

I'm testing it here: https://suzanna-linn.github.io/slua/moving-lua-definitions, check "Simple Types" at top right.

I had the idea of `type-doc` but since it can be generated by script I'm not going to add it.

I'm rather bad at type annotations, so the one that you decide.

tapple: If simple types are generated, then I'd rather, for consistency, that the types in this file would all pass typechecker, so, replace `table` with `{}`

tapple: I didn't know the binary operators supported reverse metamethods. Good catch

Suzanna-Linn: Changed 'table' to '{}'

We can do `10 * myObject` and the call is `(integer, table)`

Suzanna-Linn: But not the comparison metamethods, that need it in both operands (and with the same function)

Changed `__eq`, `__le`, `__lt`

monty-linden (approved):

PR #80, opened 2026-03-21
Luau 0.711 `math` constants

tapple:
- Release notes: https://github.com/luau-lang/luau/releases/tag/0.711
- https://github.com/luau-lang/rfcs/pull/169
- Related: #81

PR #161, opened 2026-06-24
Convert Lua's python data structure from list to dict

tapple:
- Part of #122 

A lot (all?) of the lua generators need to look up things by name, and that's easier if the internal data structure is a dict, not a list. Python-only change

The generators that need the most access to modules and classes by name are:
- #120 
- #181


lsl_definitions/generators/keywords.py
```diff
@@ -60,7 +60,7 @@ def dump_slua_syntax(
     pretty: bool = False,
```
tapple, line 60: My design thoughts regarding the structure of EmbeddedBuiltinDefinitions.cpp:

Logically, the whole module just needs to return a single string (ignoring the types library). So, it could just embed the existing secondlife.d.luau file as-is in the cpp source code

The technical reason not to is fflags.

SLua/Luau 0.710 use 2 fflags:
1. LuauTypeCheckerUdtfRenameClassToExtern
2. LuauMorePermissiveNewtableType

Both fflags only affect the types library, which this repository doesn't touch

Luau 0.723 uses 5 fflags:
1. LuauIntegerLibrary
2. LuauIntegerType2
3. LuauAllowGlobalDeclarationToBeCalledClass
4. DebugLuauUserDefinedClasses
5. LuauUdtfTypeIsSubtypeOf

fflags 1-2 affect the inclusion of the integer library and 2 buffer functions
fflags 3-4 affect the inclusion of the class library
fflags 2 and 5 affect the types library, which this repository doesn't care about

@HaroldCindy mentioned in the 2026-06-23 Server User Group that SLua is going to merge upstream from Luau soon, and that it will leave the integer and class libraries disabled by fflags for now. 
Therefore, I think it best to match the structure of the generated file as closely to upstream as possible, so that it can keep the fflags working. Is that ok?

tapple, line 60: I'm unsure which version of Luau I should target for the first EmbeddedBuiltinDefinitions generator:
1. 0.710 (SLua current)
2. 0.723 (Luau latest)
3. something else

Mostly it depends on how soon you plan to merge Luau:
1. 0.710 has all the data in slua_definitions.yaml already
2. 0.723 is missing data for the  integer library, class library, and read-only indicators

tapple, line 60: I'm also curious if you want me to put an EmbeddedBuiltinDefinitions.cpp generator here or in the slua repo. I know you had opinions on Rider making the dict wrappers here vs slua repo. I'll put it in lsl-definitions repo unless you say otherwise. It can be moved later if needed

tapple: I've gone ahead and targeted Luau 0.729 for `EmbeddedBuiltinDefinitions.cpp`: #181 
I made the following Luau update PR's in preparation:
- #80 
- #95 
- #178
- #165
- #176
- #179

tapple: moved to #181

tapple: moved to #181

tapple: moved to #181

monty-linden (approved):
lgtm

PR #163, opened 2026-06-25
Add slua_definitions.yaml to generated output

### LINKED ISSUE #162: Add a completed `slua_definitions.yaml` to generated output.
Currently several parts of `slua_definitions` are generated from `lsl_definitions` internally by the python code.

See `# auto-generated from lsl_definitions.yaml` in https://github.com/secondlife/lsl-definitions/blob/main/slua_definitions.yaml

It would be useful to have the complete data output to a file in generated.

Then that file be served by the sim capabilities instead of the incomplete one.

WolfGangS:
Add a version of `slua_definitions` to the generated output, that has all the content generated from lsl_defintitions included.

Partially resolves #162 Would still need LL implementation to have this file be served in the capabilities instead of the incomplete one.

### DIFF
--- a/lsl_definitions/generators/slua.py
+++ b/lsl_definitions/generators/slua.py
@@ -256,3 +256,36 @@ def gen_selene_yml(definitions: LSLDefinitions, slua_definitions: SLuaDefinition
     noalias_dumper.ignore_aliases = lambda self, data: True
     yaml.dump(selene, file, sort_keys=False, Dumper=noalias_dumper)
     return file.getvalue()
+
+
+@register("slua_definitions")
+def gen_slua_definitions(definitions: LSLDefinitions, slua_definitions: SLuaDefinitions) -> str:
+    """Generate SLua definitions file"""
+    quaternion_module = slua_definitions.get_module("quaternion")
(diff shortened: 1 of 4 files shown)


lsl_definitions/generators/slua.py
```diff
@@ -256,3 +256,36 @@ def selene_module(module: SLuaModule) -> dict:
...
+    """Generate SLua definitions file"""
+    quaternion_module = slua_definitions.get_module("quaternion")
+    rotation_module = SLuaModule(
+        name="rotation",
+        comment=quaternion_module.comment,
+        callable=quaternion_module.callable,
+        constants=quaternion_module.constants,
+        functions=quaternion_module.functions,
+    )
+    slua_definitions.modules.append(rotation_module)
```
WolfGangS, line 272: I suspect this should be moved to the main method that builds up slua_definitions, and not done here, that would also need to be changed in the textmate generator too then.

tapple: maybe. only the d.luau generator does anything with the rotation in globalVariables in the raw file. Every other handler does duplicate it

PR #33, opened 2025-12-09
`bool-semantics` for function parameters

WolfGangS:
This adds the `bool-semantics` property for function parameters if they are integer values that operate with a `0 - off` else `on` mechanic.

Resolves #32

I have added the needed yamale changes for schema validation. If #30 is completed before or after this adjustments will be needed in either feature.

PR #164, opened 2026-06-27
Reusable LSL type semantics

### LINKED ISSUE #31: Proposal: Constant definitions - grouping, subtypes, arguments
Would be good to see constants have _grouping_ and whether they are _bit mask_, _bit field_ or _enum_.

While most constants groups can be inferred by their prefix, e.g. `PRIM_`, there are others such as the sensor/type constants that did not have this foresight (or deprecation to prefixed versions).

Additionally there are MANY constants used with _arguments_ style syntax in their tooltips, this should be integrated as a dedicated field. For example:
```yaml
  REZ_POS:
    tooltip: Position at which to rez the new object.
    type: integer
    subtype: enum
    value: '2'
    grouping: Rezzing
    arguments: '[vector position, integer relative, integer atroot]]'
```

There are however certain contexts based on _setting_ or _getting_ for the constants, particularly with the prim params with `llGetLinkPrimitiveParams` & `llSetLinkPrimitiveParamsFast`.

Alternatively we could call it _usage_ instead of _arguments_, that might be more towards how it is meant to be used for documentation purposes with tooltips and like. Here's how that might look:
```yaml
  PRIM_TEXTURE:
    tooltip: 'Sets or gets the Blinn-Phong texture parameters on the prim'
    type: integer
    subtype: enum
    value: 17
    grouping: 'Prim Params'
    usage:
      setter: '[integer face, string texture, vector repeats, vector offsets, float rotation_in_radians]'
      getter: '[integer face]'
```

Multiple groupings:
```yaml
  AGENT:
    tooltip: Objects in world that are agents.
    type: integer
    subtype: bitfield
    value: '0x1'
    grouping:
    - Sensor
    - Type
  ACTIVE:
    tooltip: Objects in world that are running a script or currently physically moving.
    type: integer
    subtype: bitfield
    value: '0x2'
    grouping:
    - Sensor
    - Type
  PASSIVE:
    tooltip: Static in-world objects.
    type: integer
    subtype: bitfield
    value: '0x4'
    grouping:
    - Sensor
    - Type
  SCRIPTED:
    tooltip: Scripted in-world objects.
    type: integer
    subtype: bitfield
    value: '0x8'
    grouping:
    - Sensor
    - Type
```

Thoughts?

tapple:
`lsl_definitions.yaml` has `-semantics` fields that tightly associate with the type field. This PR combines them into a shared data structure. This merges how rulesets and functions were handling type semantics into one reusable data structure.

I also verified that slua's `gen_primparams_methods.py` requires no code changes and generates an identical `llprim_set_primitive_params.inl` despite the type class changing from `BuilderParamType` to `LSLTypeSemantics`. 

I expect the rulesets code change is the most contentious part of this PR; it is isolated to a single commit just in case: https://github.com/secondlife/lsl-definitions/pull/164/changes/1a42af53288dea039df04ac4a8a4888ddfd91df1

Refactors data added in the following PR's
- #33 
- #51 
- #108 
- #143 
- #113
- #182 
- #183
- #177 

#182 and #183 were merged with almost zero review (partially by mistake), so this PR also serves as my review for them.

Due to the above (primarily #182), this PR is marked:
- Closes #31


Suzanna-Linn:
It looks very well.

lsl_definitions/lsl.py
```diff
@@ -246,29 +236,98 @@ class LSLEnum:
...
+    VECTOR = "vector", LSLType.VECTOR, "vector"
+    ROTATION = "rotation", LSLType.ROTATION, "quaternion"
+    LIST = "list", LSLType.LIST, "list"
+
+    INDEX = "index", LSLType.INTEGER, "number"
+    BOOLEAN = "boolean", LSLType.INTEGER, "boolean | number"
+    ASSET = "asset", LSLType.STRING, "string | uuid"
+    ENUM = "enum", LSLType.INTEGER, "number"
+    PARAM = "param", LSLType.LIST, "{number}"
+    PARAM_GET = "param-get", LSLType.LIST, "{number}"
```
tapple, line 257: - These are from #113. I added them only for illustrative purposes. They are unused until one of these PR's is merged and the other rebased.

tapple: This and #113 have many conflicts. It would be slightly easier for me if #113 was merged first, due to this PR having no generated file changes, while #113 does. But, I can handle either order

tapple: the `ruleset` parameter from #177 should also be moved to `LSLTypeSemantics` if that gets merged before this PR

PR #167, opened 2026-06-30
Adding info to generate wiki tables (constants)

Suzanna-Linn:
This PR is built on #113.

After commenting in #160, I'm adding these fields to the constants that have associated values:
- in constants with one associated value:
  - value-type: same as #143.
  - range: array [ start, end ].
  - default: default value.
  - enum: category of constants, or group of categories.
  - details: anything else in the Wiki table.
  - access:
    - "arg": the value is always an argument.
    - "read": it's only used when reading.
    - "write": it's only used when writing.
    - "read/write": default, not used in the file.
  - values: an array when there is more than one associated value (or only one, but with an informative name):
    - name
    - value-type
    - range
    - default
    - enum
    - details
    - access

JSON file with the added fields: https://suzanna-linn.github.io/slua/wiki-constants-values.html

LSL definitions browser with the updated YAML file: https://suzanna-linn.github.io/slua/reference-lsl-definitions

After #113 and this one are merged, I will write another PR for the functions, as commented in #160.


Suzanna-Linn:
Changed to draft. It will need redoing after #113 is merged, and I have some more to add.

Issue #175, opened 2026-07-07
Merge LL table wrappers with the ll function


tapple:
SLua currently has 3 table wrappers over LSL functions:
- `llhttp.response` wraps `ll.HTTPResponse`
- `llprim.particleSystem` wraps `ll.LinkParticleSystem`
- `llprim.setMedia` wraps `ll.SetLinkMedia`

Since these are mostly LSL functions with a fresh coat of paint, they aren't truly new, and might confuse Lua users who are looking for new features. I think it would be preferable to just have the `ll` function be able to accept either the backwards-compatable list, or the new dictionary. They can be easily distinguished by whether the table keys are numbers or strings

If a caller passes a table with both array and dict parts, it should probably be an error, though, they could be merged. Passing both defeats one of the main reasons for the wrappers in the first-place: typechecking

<details><summary>Full conversation from 2026-07-07 Server User Group</summary>

> [12:33 PM] Apple Goat: llhttp.request will be in the next lua build. It's a thin wrapper over llHTTPRequest. I would like it to be removed and save the name for a much more object-oriented http library that's lua-exclusive sometime in the next few years. dunno if anyone else has an opinion
> [12:34 PM] Harold (haroldcindy): @Apple Goat You mean one that optionally uses coroutines or something like that so you can get a response object?
> [12:34 PM] Apple Goat: @Harold (haroldcindy) coroutines and request/response classes
> [12:34 PM] Harold (haroldcindy): hmmm
> [12:35 PM] Harold (haroldcindy): @Rider Linden What do you think? Unlike the others, there's lots of additional semantics we'd probably want to add to a new HTTP wrapper other than just arg handling
> [12:35 PM] Horny Nuzzle: new http wrapper 👀
> [12:36 PM] Rider Linden: I like it @Harold (haroldcindy) , but I'm not sure when we'd be able to actually do it.  It would be a rather large value of "Soon".
> [12:37 PM] Rider Linden: We could move the llhttp:request into the ll. space...
> [12:39 PM] Suzanna (suzannalinn): i like ll. space having only functions that LSL has
> [12:39 PM] Rider Linden: I think I had suggested adding llcom. But there was some resistance to that if I recall.
> [12:39 PM] Horny Nuzzle: I think llhttp. would be a fitting compromise
> [12:40 PM] Rex Cronon: i think llcomm sounds better:)
> [12:40 PM] Rider Linden: I think that is where it is now...
[...]

PR #177, opened 2026-07-10
Allow rulesets on legacy functions

Rider-Linden:
Allows the ruleset tables to be use in the legacy functions, and removes llhttp:request

Issue #175 

### related PRs
https://github.com/secondlife/slua/pull/87
https://github.com/secondlife/server/pull/2574


lsl_definitions/lsl.py
```diff
@@ -260,8 +260,10 @@ class LSLArgument:
     """Represents an asset name or uuid"""
     bool_semantics: bool
     """Represents a boolean"""
+    ruleset: str | None
+    """Name of a builder-ruleset for dict-to-list coercion"""
```
tapple, line 280: nice. I'll need to add this to the type-semantics dict in #164

generated/secondlife_selene.yml
```diff
@@ -7986,7 +7986,7 @@ globals:
   ll.HTTPRequest:
     args:
     - type: string
-    - type: table
+    - type: any
```
tapple: in the `selene_type` function: https://github.com/secondlife/lsl-definitions/blob/8fd24201451dd853ca4e01a02581ca6b3c257919/lsl_definitions/generators/slua.py#L132-L136

add a rule 
```python
        if type_str.startswith("list | "):
            return "table"
```
so this stays a table.

Maybe I should move some of that selene logic to `LSLTypeSemantics` in #164

tapple (approved):
The generated files look right. I can't really check anything else. Approved

HaroldCindy (approved):
Maybe verb-ify `llprim.particleSystem()` -> `llprim.setParticleSystem()`? Other than that lgtm.

PR #176, opened 2026-07-08
Luau 0.715 integer library

tapple:
- Part of #122 
- Release notes: https://github.com/luau-lang/luau/releases/tag/0.715
- Documentation: https://rfcs.luau.org/type-long-integer.html
- TextMate changes submitted upstream: https://github.com/JohnnyMorganz/Luau.tmLanguage/pull/30
- More TextMate changes submitted upstream: https://github.com/JohnnyMorganz/Luau.tmLanguage/pull/31
- Uses updated bit32 and vector documentation from #172

PR #180, opened 2026-07-11
Exclude `integer` and `class` from generated files

tapple:
`integer` and `class` are part of luau, but, for the moment, they are disabled in slua by fflags. Mark them as such and exclude them from all generated files except `EmbeddedBuiltinDefinitions.cpp`. This also makes the generated files backwards-compatable with Luau 0.710

- Part of #122 
- Includes #179 
- Includes #176


generated/syntax/slua.tmLanguage
```diff
@@ -488,19 +488,19 @@
             <key>name</key>
             <string>constant.numeric.hex.luau</string>
             <key>match</key>
-            <string>\b0_*[xX]_*[\da-fA-F_]*(?:[eE][\+\-]?_*\d[\d_]*(?:\.[\d_]*)?)?</string>
+            <string>\b0_*[xX]_*[\da-fA-F_]*(?:i|[eE][\+\-]?_*\d[\d_]*(?:\.[\d_]*)?)?</string>
```
tapple, line 491: The only integer/class change this doesn't exclude is the `4i` syntax in the markdown file, as it was added in the template and not in the generator. But, I think that's the most harmless part of the change

PR #181, opened 2026-07-11
Generate `EmbeddedBuiltinDefinitions.cpp`

### LINKED ISSUE #122: Create a generator for `EmbeddedBuiltinDefinitions.cpp`

tapple:
Generate `EmbeddedBuiltinDefinitions.cpp` so that `luau-analyze` knows about `ll` functions and such.

This generates a file intended for Luau 0.725 - 0.730

- Closes #122
- https://github.com/tapple/lsl-definitions/pull/3


templates/EmbeddedBuiltinDefinitions.in.cpp
```diff
@@ -0,0 +1,523 @@
+// This file is part of the Luau programming language and is licensed under MIT License; see LICENSE.txt for details
+#include "Luau/BuiltinDefinitions.h"
```
tapple, line 2: My design thoughts regarding the structure of EmbeddedBuiltinDefinitions.cpp:

Logically, the whole module just needs to return a single string (ignoring the types library). So, it could just embed the existing secondlife.d.luau file as-is in the cpp source code

The technical reason not to is fflags.

- [SLua's version](https://github.com/secondlife/slua/blob/main/Analysis/src/EmbeddedBuiltinDefinitions.cpp) (Luau 0.710)
- [Luau's version](https://github.com/luau-lang/luau/blob/master/Analysis/src/EmbeddedBuiltinDefinitions.cpp) (Luau 0.725)

SLua/Luau 0.710 use 2 fflags:
1. LuauTypeCheckerUdtfRenameClassToExtern
2. LuauMorePermissiveNewtableType

Both fflags only affect the types library, which this repository doesn't touch

Luau 0.723 uses 5 fflags:
1. LuauIntegerLibrary
2. LuauIntegerType2
3. LuauAllowGlobalDeclarationToBeCalledClass
4. DebugLuauUserDefinedClasses
5. LuauUdtfTypeIsSubtypeOf

fflags 1-2 affect the inclusion of the integer library and 2 buffer functions
fflags 3-4 affect the inclusion of the class library
fflags 2 and 5 affect the types library, which this repository doesn't care about

@HaroldCindy mentioned in the 2026-06-23 Server User Group that SLua is going to merge upstream from Luau soon, and that it will leave the integer and class libraries disabled by fflags for now. 
Therefore, I think it best to match the structure of the generated file as closely to upstream as possible, so that it can keep the fflags working. Is that ok?

tapple, line 2: I'm unsure which version of Luau I should target for the first EmbeddedBuiltinDefinitions generator:
1. 0.710 (SLua current)
2. 0.723 (Luau latest)
3. something else

Mostly it depends on how soon you plan to merge Luau:
1. 0.710 has all the data in slua_definitions.yaml already
2. 0.723 is missing data for the  integer library, class library, and read-only indicators

tapple: I've gone ahead and targeted Luau 0.729 for `EmbeddedBuiltinDefinitions.cpp`: #181 
I made the following Luau update PR's in preparation:
- #80 
- #95 
- #178
- #165
- #176
- #179

tapple, line 2: I'm also curious if you want me to put an EmbeddedBuiltinDefinitions.cpp generator here or in the slua repo. I know you had opinions on Rider making the dict wrappers here vs slua repo. I'll put it in lsl-definitions repo unless you say otherwise. It can be moved later if needed

tapple: I chose to generate a file for any of Luau 725, 726, 727, 728, or 729 because:
-  `EmbeddedBuiltinDefinitions.cpp` is identical in those 5 versions
- it has minimal fflags that my generator needs to worry about (especially no fflags that switch between 2 definitions of one function, which the generator doesn't support
- The file has been stable for quite a few versions (5), but there's upcoming PR's that will add more FFlags to worry about:
  - https://github.com/luau-lang/luau/pull/2337
  - https://github.com/luau-lang/luau/pull/2378
  - https://github.com/luau-lang/luau/pull/2540

lsl_definitions/generators/slua_embedded.py
```diff
@@ -0,0 +1,71 @@
...
+        inserts["DEBUG_TABLE"] = defs.getvalue()
+    with io.StringIO() as defs:
+        slua_definitions.modules.pop("utf8").write_luau_def(defs)
+        inserts["UTF8_TABLE"] = defs.getvalue()
+    with io.StringIO() as defs:
+        slua_definitions.modules["buffer"].write_luau_def(defs)
+        inserts["BUFFER_TABLE_NOINTEGER"] = defs.getvalue()
+    with io.StringIO() as defs:
+        slua_definitions.modules.pop("buffer").write_luau_def(defs, enable_fflags=True)
+        inserts["BUFFER_TABLE"] = defs.getvalue()
```
tapple, line 67: yunno, it's not really sustainable trying to generate 2 different versions of a definition for different `FFlag`s.

I think a better idea would be to:
1. Only generate the parts of `EmbeddedBuiltinDefinitions.cpp` that SLua actually enables with it's released `FFlag` set
2. Leave the `FFlag`-disabled parts of `EmbeddedBuiltinDefinitions.cpp` in the template, to be synced upstream from Luau

In practice, for Luau 0.729, this would mean:
1. generating `kBuiltinDefinitionBufferSrc_NOINTEGER`
2. NOT generating `kBuiltinDefinitionBufferSrc`
2. NOT generating `kBuiltinDefinitionIntegerSrc`
2. NOT generating `kBuiltinDefinitionClassSrc`
3. Removing #176, #179, and #180 as dependencies on this PR
4. Closing #180 because it's no longer needed (it's what implements `enable_fflags` in the commented code)
5. Leaving #176 and #179 unmerged until they are actually enabled via `FFlag`s in SLua, and not just present and disabled in the code

This would also mean that I don't need to ask Harold not to base SLua on a particular Luau version just because of what FFlags Luau happened to put in `EmbeddedBuiltinDefinitions.cpp` that week (see above comment https://github.com/secondlife/lsl-definitions/pull/181#discussion_r3572340039)

templates/EmbeddedBuiltinDefinitions.in.cpp
Rider-Linden: I generate the wrappers here and include them in the actual simulator since putting them into the VM would require a new VM every time we added a flag to any function.  (I have the same opinion about prim params, but it is not a hill I'm willing to die on.)

We provide the generators, and the pregenerated .cpp files in the distribution so 3rd parties can include them easily.

gen_all_definitions.sh
```diff
@@ -42,6 +42,7 @@ $CMD $DEFS gen_tree_source_file "$outdir/cpp/lscript_tree.cpp"
 $CMD $DEFS gen_lexer_file ./templates/indra.in.l "$outdir/templated/indra.l"
 $CMD $DEFS gen_parser_file ./templates/indra.in.y "$outdir/templated/indra.y"
 $CMD $DEFS gen_mono_library_defs ./templates/LslLibrary.cs "$outdir/templated/LslLibrary.cs"
+$CMD $DEFS gen_slua_embedded_defs ./templates/EmbeddedBuiltinDefinitions.in.cpp $SLUA "$outdir/cpp/EmbeddedBuiltinDefinitions.cpp"
```
tapple, line 45: - I may close this PR as lute just gained definitions-loading capabilities, and that sounds more sustainable than this approach: https://github.com/luau-lang/lute/pull/1234

PR #184, opened 2026-07-21
[RFC] Adding tables without constants

Suzanna-Linn:
Adding enums of tables that haven't got constants defined.

SimulatorDataRating and SimulatorDataStatus are used in #185.

### DIFF
--- a/lsl_definitions.yaml
+++ b/lsl_definitions.yaml
@@ -555,6 +555,416 @@ enums:
     type: enum
     prefix: VEHICLE_
     tooltip: ''
+  AnimationStates:
+    type: enum
+    tooltip: Names of the animation states
+    value-type: string
+    other-values: true
+    members:
+      - Crouching:
+          tooltip: ""
+          details: "Type: State, Default: crouch"
+      - CrouchWalking:
+          tooltip: ""
+          details: "Type: State, Default: crouchwalk"
+      - Falling Down:
+          tooltip: ""
+          details: "Type: State, Default: falldown"
+      - Flying:
+          tooltip: ""
+          details: "Type: State, Default: fly"
+      - FlyingSlow:
+          tooltip: Transition between hovering and forward flight.
+          details: "Type: State, Default: flyslow"
+      - Hovering:
+          tooltip: ""
+          details: "Type: State, Default: hover"
+      - Hovering Down:
+          tooltip: ""
+          details: "Type: State, Default: hover_down"
+      - Hovering Up:
+          tooltip: ""
+          details: "Type: State, Default: hover_up"
+      - Jumping:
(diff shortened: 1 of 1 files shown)


lsl_definitions.yaml
```diff
@@ -555,6 +555,416 @@ enums:
     type: enum
     prefix: VEHICLE_
     tooltip: ''
+  AnimationStates:
+    type: enum
+    tooltip: Names of the animation states
+    value-type: string
+    other-values: true
```
tapple: no other values

lsl_definitions.yaml
```diff
@@ -555,6 +555,416 @@ enums:
     type: enum
     prefix: VEHICLE_
     tooltip: ''
+  AnimationStates:
```
tapple: Sort the new enums alphabetically into the existing enums

lsl_definitions.yaml
```diff
@@ -555,6 +555,416 @@ enums:
...
+          details: "Typecast to: integer boolean"
+      - region_product_name:
+          tooltip: 'Get the type of region this is: "Estate / Full Region", "Mainland /
+            Homestead", "Estate / Openspace", "Estate / Full Region - Skill
+            Gaming" etc.'
+      - region_product_sku:
+          tooltip: Get the region's product number (as string).
+      - region_start_time:
+          tooltip: Get the time the region was last (re)started, in llGetUnixTime format
+          details: "Typecast to: integer"
```
tapple: What's the purpose of having both "tooltip" and "details"? I feel like they could be combined, unless you have specific purposes in mind for each one

lsl_definitions.yaml
tapple: "ALL" should be another documented value. then you won't need other values

Suzanna-Linn: The main reason to add "other-values" is the `llGetAnimation()` caveat in Wiki:
- "New return values could conceivably be added at any time and this list may not in-fact be complete. Scripts should be written under the assumption that they may receive a value they won't recognize."

Also `llGetAnimation()` has two more tables with possible values.

"ALL" is only used in `llResetAnimationOverride()`.

lsl_definitions.yaml
Suzanna-Linn: Sorted (also sorted some others as a side effect).

lsl_definitions.yaml
Suzanna-Linn: - "tooltip" is the tooltip itself.
- "details" has the other columns in the Wiki tables, formatted as "column name: cell content, ...", also used in `OBJECT_*` and `PRIM_*`.

Using two fields to preserve the Wiki structure and because they have different format.

PR #185, opened 2026-07-21
[RFC] Adding dataserver return data

Suzanna-Linn:
Adding dataserver return data for the asynchronous functions.

SimulatorDataRating and SimulatorDataStatus are defined in #184.

### DIFF
--- a/lsl_definitions.yaml
+++ b/lsl_definitions.yaml
@@ -9396,16 +9396,6 @@ functions:
     experience: true
     func-id: 387
     return: key
-    sleep: 0.0
-    tooltip: Starts an asynchronous transaction to create a key-value pair (k and v)
-      associated with the script's experience. Returns a key query handle for
-      the dataserver event. Fails with XP_ERROR_STORAGE_EXCEPTION if the key
-      already exists.
-    categories:
-      - data_storage
-      - dataserver
-      - experience
-      - experience_data
     dataserver-semantics:
       value-type: string-csv
       error-semantics: true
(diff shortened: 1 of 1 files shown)


lsl_definitions.yaml
```diff
@@ -9431,6 +9477,22 @@ functions:
       - dataserver
       - experience
       - experience_data
+    dataserver-semantics:
```
tapple, line 9471: sort `dataserver-semantics` higher so it's right below `return`. Leave tooltip and category at the bottom, so that it's next to all the other semantics

lsl_definitions.yaml
```diff
@@ -1719,6 +1733,13 @@ constants:
...
+    deprecated:
+      reason: Ratings were removed from SL.
+    dataserver-semantics:
+      value-type: string-csv
+      tooltip: "Returns [0, 0, 0, 0, 0, 0]. Used to return: [pos_behavior,
+        neg_behavior, pos_appearance, neg_appearance, pos_building,
+        neg_building]."
   DATA_RESERVED_0:
     member-of: [AgentData]
     tooltip: Reserved for Linden use.
```
tapple, line 1745: I wonder if this actually even belongs to `AgentData` (I just guessed it did from it's name). It doesn't generate a dataserver event when I pass `DATA_RESERVED_0` to `llRequestAgentData` (requires a Lua sim)

lsl_definitions.yaml
Suzanna-Linn: Moved.

lsl_definitions.yaml
Suzanna-Linn: No idea... Let's wait for a Linden...

PR #188, opened 2026-08-28
Ruleset: llSetKeyframedMotion

tapple:
Add a table ruleset for the options parameter of:
- [ll.SetKeyframedMotion](https://wiki.secondlife.com/wiki/LlSetKeyframedMotion)
- llprim.keyframedMotion

No change to the keyframes parameter


Rider-Linden (approved):
Looks good to me.

HaroldCindy:
Is there a strong case for having a separate function as well, if the existing `ll.SetKeyframedMotion` has to handle it as well? Does it improve the typing situation by much?

tapple:
> Is there a strong case for having a separate function as well, if the existing `ll.SetKeyframedMotion` has to handle it as well? Does it improve the typing situation by much?

No. deleted llprim.keyframedMotion

PR #192, opened 2026-09-08
Deprecate `ll.IntegerToBase64`

tapple:
I seem to have missed this function when I deprecated the rest of the lsl Base64 functions
- Followup to #76

I verified the recommended improvement is a match with this script:
```luau
for i = -0x8000_0000, 0x7FFF_FFFF, 4321 do
    local lsl = ll.IntegerToBase64(i)
    local lua = llbase64.encode(string.pack(">i", i))
    if (lsl ~= lua) then print("mismatch", i) end
end
print(ll.GetScriptName(), "done")
```
`string.pack` is about half as slow as `ll.IntegerToBase64`:
```luau
for i = -0x8000_0000, 0x7FFF_FFFF, 4321 do
    local lua = llbase64.encode(string.pack(">i", i))
end
print("string.pack done in", os.clock()) --> 117.44394799995199
```
```luau
for i = -0x8000_0000, 0x7FFF_FFFF, 4321 do
    local lsl = ll.IntegerToBase64(i)
end
print("lsl done in", os.clock()) --> 62.55046899997564
```
But buffers are very slightly faster than `ll.IntegerToBase64`
```luau
local b = buffer.create(4)
for i = -0x8000_0000, 0x7FFF_FFFF, 4321 do
    buffer.writei32(b, 0, bit32.byteswap(i))
    local lua = llbase64.encode(b)
end
print("buffer done in", os.clock()) --> 62.09301100001903
```


HaroldCindy (approved):
Good riddance!

PR #193, opened 2026-09-10
Export `JsonOptions` types

tapple:
The typechecker likes it better when I explicitly type these, but I can't as they aren't exported


HaroldCindy (approved):
LGTM, thanks!

PR #197, opened 2026-09-20
Improve `table.create` type

tapple:
Previously, this code would fail typechecking under the new solver:
```luau
local points: {number} = table.create(5)
```
```
TypeError: Expected this to be '{number}', but got '{unknown}'; 
Expected the indexer result to be exactly `number`, but got `unknown`
```
Now it gives no error

The old solver gave no error before, and still doesn't


HaroldCindy (approved):
Makes sense, thanks!

